PR #9, opened 2015-11-05
Added docs for Linux (Debian) cross compiler

brendank310:

### DIFF
--- a/doc/cross_compilers/linux_cross_compiler.md
+++ b/doc/cross_compilers/linux_cross_compiler.md
@@ -1,0 +1,91 @@
+# Linux Cross Compiler (GCC)
+
+## Summary
+
+The following describes how to create a cross-compiler for the Bareflank hypervisor on Linux. A script to automate this is provided in the repository.
+
+## Apt
+
+This guide was designed around a bare install of Debian 8.0 (Jessie). Prerequisites for building the cross compiler are satisfied using apt.
+
+[Apt wiki](https://wiki.debian.org/Apt)
+
+You will need to install the following:
+
+```
+apt-get install build-essentials
+apt-get install libmpc
+```
+
+## Downloads
+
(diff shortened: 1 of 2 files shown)


rianquinn:
+1

doc/cross_compilers/linux_cross_compiler.md
```diff
@@ -0,0 +1,91 @@
...
+
+This guide was designed around a bare install of Debian 8.0 (Jessie). Prerequisites for building the cross compiler are satisfied using apt.
+
+[Apt wiki](https://wiki.debian.org/Apt)
+
+You will need to install the following:
+
+```
+apt-get install build-essentials
+apt-get install libmpc
```
rianquinn: Does this provide ISL?

brendank310: No it doesn't. What does ISL provide that is necessary? Everything seems to function just fine on Linux without it (no failed unittests).

rianquinn: It's used to perform a lot of the optimizations. You can build without it, but if we ever want to turn on optimizations (like loop unrolling), ISL is needed. Figured might as well get them included in the instructions now rather than later. For Homebrew and Cygwin, it was as simple as installing libisl-dev (also might come up as cloog).

rianquinn: When you run the configure script for GCC, you can validate if ISL is being used as it will tell you. It's grouped up with the checks for MPC, MPFR, etc...

tools/scripts/linux-bareflank-cross-compiler.sh
```diff
@@ -0,0 +1,38 @@
...
+export GCC_TARGET=x86_64-elf
+export BINUTILS_TARGET=x86_64-elf
+export PATH="$PREFIX/bin:$PATH"
+
+apt-get install build-essentials libmpc-dev libisl-dev
+
+mkdir -p $PREFIX
+
+pushd ~/Downloads
+#wget https://ftp.gnu.org/gnu/gcc/gcc-5.2.0/gcc-5.2.0.tar.bz2
```
rianquinn: would it be possible to define these as variables that can be overridden if we want to try a different version? Would be useful for Travis CI

Issue #15, opened 2015-11-08
Compilation warnings with Linux


rianquinn:
When compiling on Ubuntu LTS server 12.04, you get the following warnings:

```
gcc elf_loader.c -o .build/elf_loader.o -c  -fpic -DENABLE_ELF_DEBUGGING -MT .build/elf_loader.o -MMD -MP -MF .build/elf_loader.Td -I./ -I../include/
elf_loader.c: In function ‘elf_file_print_header’:
elf_loader.c:693:5: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_addr’ [-Wformat]
elf_loader.c:694:5: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_off’ [-Wformat]
elf_loader.c:694:5: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 4 has type ‘elf64_off’ [-Wformat]
elf_loader.c:695:5: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_off’ [-Wformat]
elf_loader.c:695:5: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 4 has type ‘elf64_off’ [-Wformat]
elf_loader.c: In function ‘elf_print_section_header’:
elf_loader.c:880:5: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_addr’ [-Wformat]
elf_loader.c:881:5: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_off’ [-Wformat]
elf_loader.c:882:5: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_xword’ [-Wformat]
elf_loader.c:885:5: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_xword’ [-Wformat]
elf_loader.c:886:5: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘elf64_xword’ [-Wformat]
elf_loader.c: In function ‘elf_print_sym’:
[...]

PR #16, opened 2015-11-09
Fix compilation warnings with GCC on Linux

rianquinn:
GCC doesn't understand how to use printf with the standard
integer types. To get them to work, you have to use some
macros, which this patch uses instead.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/15


brendank310:
Personally don't agree with moving doxygen data from the headers to the sources but the inttypes change is fine.

rianquinn:
Will move the API docs to the .h, and update the Governance Model in a separate (soon) pull request

Issue #17, opened 2015-11-09
GCC does not recognize -std=c++11


rianquinn:
Unless you have GCC >= 4.7, it doesn't recognize -std=c++11. Instead, you need to use -std=c++0x

PR #18, opened 2015-11-09
Fix issue with GCC and c++11 features

rianquinn:
GCC 4.6 (used by Travis CI) does not recognize c++11. Instead
it uses c++0x, which does not support the "override" keyword.
As a result, the following patches remove c++11 to support
Travis CI, as well as the override command.

Note: These changes are only needed for the native compiler.
Cross-compiled code will use GCC 5.2 and above which has
support for c++14.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/17

### DIFF
--- a/elf_loader/test/Makefile
+++ b/elf_loader/test/Makefile
@@ -30,7 +30,7 @@ RM=rm -rf
 MD=mkdir -p
 
 CCFLAGS=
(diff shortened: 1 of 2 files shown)


brendank310:
LGTM

Issue #21, opened 2015-11-09
Windows unit test failing


rianquinn:
The ELF loader does not compile under Windows, and the resulting binary does not execute correctly under Windows. 

The compilation errors include:

```
g++ test.cpp -o .build/test.o -c -std=c++0x  -MT .build/test.o -MMD -MP -MF .build/test.Td -I../include/ -I./ -I../../include/
In file included from ./test.h:25:0,
                 from test.cpp:22:
../../include/unittest.h: In member function ‘virtual int unittest::run()’:
../../include/unittest.h:53:20: error: ‘EXIT_FAILURE’ was not declared in this scope
             return EXIT_FAILURE;
                    ^
../../include/unittest.h:58:20: error: ‘EXIT_FAILURE’ was not declared in this scope
             return EXIT_FAILURE;
                    ^
../../include/unittest.h:66:24: error: ‘EXIT_FAILURE’ was not declared in this scope
                 return EXIT_FAILURE;
                        ^
../../include/unittest.h:76:20: error: ‘EXIT_FAILURE’ was not declared in this scope
             return EXIT_FAILURE;
                    ^
../../include/unittest.h:80:20: error: ‘EXIT_FAILURE’ was not declared in this scope
             return EXIT_FAILURE;
                    ^
../../include/unittest.h:82:16: error: ‘EXIT_SUCCESS’ was not declared in this scope
         return EXIT_SUCCESS;
                ^
../../common/common_target.mk:119: recipe for target '.build/test.o' failed
make[2]: *** [.build/test.o] Error 1
../common/common_subdir.mk:38: recipe for target 'all' failed
make[1]: *** [all] Error 2
common/common_subdir.mk:38: recipe for target 'all' failed
make: *** [all] Error 2
```

PR #22, opened 2015-11-11
Fix issue with Windows support

rianquinn:
Currently, Windows doesn't support the ELF loader. Other than
some missing headers, the biggest issue is Windows use of
MS x64 calling convention that prevents Windows code from
executing a symbol in the cross-compiled code. This patch
allows the Windows code to not only build, but it also provides
support for it to execute System V ABI compiled symbols.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/21

### DIFF
--- a/common/abi_conversion.asm
+++ b/common/abi_conversion.asm
@@ -1,0 +1,41 @@
+global exec_ms64tosv64
+
+section .text
+exec_ms64tosv64:
+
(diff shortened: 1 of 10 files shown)


brendank310:
Looks good to me, don't have the cycles right now to give a run through.

PR #26, opened 2015-11-25
Scafolding for Bareflank Manager

rianquinn:
This patch contains:
- The scafolding for the bareflank manager (the frontend for
  starting and stopping the hypervisor)
- Cleanup for the unit tests
- Cleanup for the ELF Loader
- Upgrades to use C++14
- Addition of Hippo Mocks for unit testing


bfm/include/ioctl_driver.h
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class ioctl_driver
+{
+public:
+
+    ioctl_driver(const ioctl& ctl, const command_line_parser& clp);
+    virtual ~ioctl_driver();
+
+    virtual ioctl_driver_error::type
```
brendank310: extraneous whitespace

rianquinn: ?

brendank310: should it not be:
virtual ioctol_driver_error::type process() const;

rianquinn: No. Return type is on it's own line. The only time I have not been doing that is in the unittest .h files with "void", and that's really only because it does make the unit tests easier to read. Otherwise you would have a lot of:

```
void
test_blah
void
test_blah2
void
test_blah3
```

brendank310: I both agree and disagree. Don't like it in the header, like it in the implementation.

rianquinn: I am fine with that so long as the function definition still moves to the next line if it starts getting too long. Once it does move to the next line, it should have one line per arg to keep it clean. Also... we will have to update the elf loader to match

PR #27, opened 2015-11-27
Initial Linux Driver Entry

rianquinn:
This patch provides the initial Linux driver entry point code.
The IOCTLs have not been implemented yet, but the driver itself
is capable of being communicated with. The API is also defined.


driver_entry/src/arch/linux/entry.c
```diff
@@ -0,0 +1,93 @@
...
+{
+    int ret;
+
+    /*
+     * TODO: At some point we should use the newer API that automatically
+     *       creates a MAJOR number for us. To get that to work, we need
+     *       to get udev working as well to create our /dev/bareflank noce
+     */
+
+    if((ret = register_chrdev(BAREFLANK_MAJOR, BAREFLANK_NAME, &fops)) < 0)
```
brendank310: I think misc_device may be more appropriate. Make the dev node much easier.

PR #31, opened 2015-12-02
New Build System / Debug Ring

rianquinn:
This patch contains a new build system. The VMM code actually
needs to be compiled twice. Once for the VMM itself using the
cross compiler, and once for the unit tests which need to
use the native compilers. The previous build system didn't
clearly define the difference.

This patch also contains the scafolding for a basic debugging
library for the VMM to demonstrate how to layout the different
files a class will need in the VMM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,12 @@ install:
 
 script:
 - make
-- cd bfelf_loader/bin
+- cd bfelf_loader/bin/native/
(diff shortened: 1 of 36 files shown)


brendank310:
Mostly looks good, though I see a lot of astyle related changes. I'd prefer those changes don't come up frequently like they having past. Perhaps we can add a test to travis-ci to checkout the repo, run astyle see that the git diff is empty.

PR #32, opened 2015-12-04
BFM Dump Support

rianquinn:
This patch adds support for dumping the debug ring
from the VMM. While attempting to add support for this code
a lot of mods were made to cleanup the C++ environment as
GCC adds symbols to shared libraries once things like
inheritance are used.

Also added support for static libraries for the unit tests.
This will enable the use of HippoMocks for mocking up C
functions (as shared libraries cannot be used in this case).

Finally sevral unit tests were either added or cleaned up
in this patch, with further changes still to come


brendank310:
Break up your pull requests!

PR #33, opened 2015-12-04
[serial] Initial commit of some of the serial port handling bits.

brendank310:


bfvmm/src/serial/include/bf_portio.h
```diff
@@ -0,0 +1,11 @@
+#ifndef BF_PORTIO__H
+#define BF_PORTIO__H
+
+extern "C" {
```
rianquinn: I would wrap these with the cpp macros as a safe guard (just in case they need to be used by C in the future)

bfvmm/src/serial/include/serial_port_x86.h
```diff
@@ -0,0 +1,210 @@
...
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#ifndef SERIAL_X86__H
+#define SERIAL_X86__H
+
+#include <port_io.h>
+#include <serial_port.h>
+/**
```
rianquinn: For C++ use ///

bfvmm/src/serial/include/serial_port_x86.h
```diff
@@ -0,0 +1,210 @@
...
+
+struct __data_bits
+{
+    uint8_t data_size;
+    uint8_t data_bits;
+};
+
+extern struct __data_bits data_size_table[];
+
+/* Receive buffer register */
```
rianquinn: Same thing here, I wouldn't mix C and C++ comments styles.

bfvmm/src/serial/src/Makefile
```diff
@@ -0,0 +1,90 @@
...
+# Cross Flags
+################################################################################
+
+CROSS_CC=~/opt/cross/bin/x86_64-elf-gcc
+CROSS_CXX=~/opt/cross/bin/x86_64-elf-g++
+CROSS_ASM=~/opt/cross/bin/nasm
+CROSS_LD=~/opt/cross/bin/x86_64-elf-ld
+
+CROSS_CCFLAGS=
+CROSS_CXXFLAGS=-c
```
rianquinn: I don't think that "-c" is needed. If it is, there is something wrong with the common_target.mk

bfvmm/src/serial/src/Makefile
```diff
@@ -0,0 +1,90 @@
...
+CROSS_LDFLAGS=-fPIC
```
rianquinn: -fPIC in LDFLAGS would be ignored. fPIC only makes sense while compiling. Actually I think its ignored by LD. fPIC is being applied by common_target.mk automatically if you specify TARGET_TYPE=lib or staticlib

bfvmm/src/serial/src/Makefile
```diff
@@ -0,0 +1,90 @@
...
+# Native Flags
+################################################################################
+
+CC=gcc
+CXX=g++
+LD=g++
+ASM=nasm
+
+CCFLAGS=
+CXXFLAGS=-c
```
rianquinn: Samething here

bfvmm/src/serial/src/Makefile
```diff
@@ -0,0 +1,90 @@
...
+
+################################################################################
+# Sources
+################################################################################
+
+TARGET_NAME=bf_serial
+TARGET_TYPE=lib
+TARGET_COMPILER=both
+
+SOURCES=serial_port_x86.cpp bf_portio.asm portio_x64.cpp
```
rianquinn: I have been converting to SOURCE+=file \n SOURCE+=file

bfvmm/src/serial/src/serial_port_x86.cpp
```diff
@@ -0,0 +1,430 @@
...
+serial_port_x86::set_parity_mode(PARITY_MODE parity)
+{
+    uint8_t i = 0;
+    uint8_t data_bits = 0;
+
+    m_parity = parity;
+
+    data_bits = m_port_io.port_read_8(m_port + LCR_OFFSET);
+
+    while (parity_table[i].mode < __PARITY_MAX)
```
rianquinn: we will likely need to get rid of the use of __ as it's a pretty big no no with C++ as GCC uses __ for reserved functions.

http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

rianquinn:
Other than those comments, looks good to me

bfvmm/src/serial/src/Makefile
brendank310: Yeah it looks like it isn't needed any longer.

bfvmm/src/serial/include/serial_port_x86.h
brendank310: I don't really agree with this. Doesn't really apply here, but I prefer longer comments use the old C style comments. Don't see a downside to having a mix.

bfvmm/src/serial/src/Makefile
brendank310: Yeah this seems to have become unnecessary following the rebase also.

bfvmm/src/serial/src/Makefile
brendank310: Not a bad idea

bfvmm/src/serial/include/serial_port_x86.h
rianquinn: I guess I'm fine with the use of C comments for large blocks, but that doesn't explain it's use for single line comments, which is clearly more difficult. But yeah, my biggest complaint with old C style comments is you cannot block comment large portions of code with C style comments everywhere. With C++ comments, you can comment, comments just fine. (granted, you can use #if 0 macros, but I just find that more of a pain

bfvmm/src/serial/src/serial_port_x86.cpp
brendank310: Woof, that's awful.

bfvmm/src/serial/src/Makefile
rianquinn: Interesting on that note.... one thing I learned is that with a makefile VAR=$VAR2 doesn't just set VAR equal to VAR2, it makes them the same thing. If VAR2 changes, so does VAR. If you use +=, that doesn't happen

bfvmm/src/serial/src/serial_port_x86.cpp
rianquinn: Yeah... basically... you cannot use "_" in C++.

PR #37, opened 2015-12-07
VMM scafolding

rianquinn:
This patch provides the initial scaffolding for the VMM.


brendank310:
Merged, github isn't detecting it though.

PR #39, opened 2015-12-09
Memory Manager Integration

rianquinn:
This patch integrates the memory manager into the VMM and
driver entry. With this patch, the VMM now has access to
a pool of pages that it can use as needed.

### DIFF
--- a/bfvmm/include/memory_manager/page.h
+++ b/bfvmm/include/memory_manager/page.h
@@ -23,6 +23,7 @@
 #define PAGE_H
 
 #include <stdint.h>
+#include <memory.h>
 
 class page
 {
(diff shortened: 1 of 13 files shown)


bfvmm/include/memory_manager/page.h
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    page();
+
+    /// Valid Page Constructor
+    ///
+    /// If given the correct values, creates a valid page
+    ///
+    /// @param phys the physical address of the page
+    /// @param virt the virtual address of the page
+    /// @param size the size of the page in bytes
```
brendank310: Probably shouldn't allow size in bytes to be specified, in the base class at least. Support for large and huge pages should probably have their own class, and the size be fixed for each class.

rianquinn: Once we create a real memory manager, we could create a page_size enum that lists the supported page sizes, and then this class could validate against that enum. It needs to be set however because the VMM will need to convert the host OS page size to the VMM's page size once that memory manager is set. For example, on Windows, we might be huge pages at 2MB, and the VMM will likely work with EPT's granularity, so it cannot be "fixed" like you see in the linux kernel.

PR #42, opened 2015-12-10
Support for VMXON

rianquinn:
This patch contains support for VMXON. A couple of notes include:
- With this patch, when you run ./bfm start vmm.modules you will
  enable VMXON and if you run ./bfm stop, you will disable VMXON
- There was a great deal of rework to remove the need for global
  objects while still maintaining seems for unit testing.
- There are no unit tests for the entry code, or VMM code.
  This code will be coming soon.

### DIFF
--- a/bfvmm/src/vmm/src/vmm_intel_x64.cpp
+++ b/bfvmm/src/vmm/src/vmm_intel_x64.cpp
@@ -19,12 +19,490 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+#include <iostream>
 #include <vmm/vmm_intel_x64.h>
 
-vmm_intel_x64::vmm_intel_x64()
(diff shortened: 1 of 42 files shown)


bfvmm/src/vmm/src/vmm_intel_x64.cpp
```diff
@@ -19,12 +19,490 @@
...
+    {
+        std::cout << "create_vmxon_region failed: "
+                  << "the allocated page is not large enough:" << std::endl
+                  << "    - page size: " << m_vmxon_page.size() << " "
+                  << "    - vmxon/vmcs region size: " << vmxon_vmcs_region_size()
+                  << std::endl;
+        return vmm_error::not_supported;
+    }
+
+    if (((uintptr_t)m_vmxon_page.phys_addr() & 0x0000000000000FFF) != 0)
```
brendank310: This could probably be a little less C like:
m_vmxon_page.phys_addr().isPageAligned() or similiar.

PR #43, opened 2015-12-10
[driver-entry] Set affinity from the kernel in Linux, not userspace.

brendank310:


rianquinn:
LTGM +1

PR #51, opened 2016-01-03
New Memory Manager

rianquinn:
The following provides a new memory manager for the VMM. With
this memory manager, it is now possible to new / delete in
the VMM. Since we do not yet have the MDL work complete, as
this will require std::array and std::hash, the VMM launch
code has been disabled for the moment. Once this code has
been complete, we will re-enable the code.


brendank310:
Just a note that a large number of tests were commented out in this PR. Is there an expected culling of these tests or was it unintentional @rianquinn?

rianquinn:
@brendank310, it was done on purpose as a lot of things have changed. Once the STL is in place, and we have the exit handler flushed out, we will have to go back to workout the unit tests for a lot of the VMM itself. Until then, since there is so much changing, it doesn't really make sense to worry about the unit tests as they are likely to change in the future.

PR #60, opened 2016-02-16
New ELF Loader

rianquinn:
The following patch provides a new ELF loader that has better
support for:

- New CTOR/DTOR logic that uses a similar approach to GCC
  with a crt static library to provide init / fini functions
- Weak symbol support
- Better searching logic
- Better unit testing (far less over-specification)
- Cleaned up formatting

In addition, this patch includes a couple of nicities that were
used when developing this code

- Comment support in the vmm.modules file
- Cleaned up build script which adds support for dependencies


brendank310:
Been testing it for the past week, works fine for all the development I've been doing.

PR #61, opened 2016-02-24
[vmm] Re-enable VMM on, VM launch, and add the ability to promote

brendank310:
a running guest to the host (VMX root mode). This allows the VMCS
to be cleared and the CPU to leave VMX root mode, while continuing
to run the guest. A generic bitmap class was added to support MSR
bitmaps in the VMM, but should be used for all such bitmaps.


.gitignore
```diff
@@ -1 +1,11 @@
 bfvmm/bin/cross/libc++.so
+.build/
+*.a
+bin/
```
rianquinn: This is a bad idea since we have a lot of files in bin/ that are in the repo

README.md
```diff
@@ -14,6 +14,9 @@ of updating the VMM logic, and re-integrating the VMCS / Exit handler
 logic. Once this is complete, we should have a working VM launch. For now, the
 VMM does not attempt a VM launch.
 
+* **2/24/2015 Update:**
```
rianquinn: If launching is working, I would just remove all of the warnings

bfvmm/include/intrinsics/intrinsics_intel_x64.h
```diff
@@ -79,8 +84,18 @@ class intrinsics_intel_x64 : public intrinsics_x64
     virtual bool vmread(uint64_t field, uint64_t *val)
     { return __vmread(field, val); }
 
+    virtual uint64_t vmread(uint64_t field)
```
rianquinn: This function should be removed. You should be checking the returned bool for an error. Likely, we will actually add the exception logic to the C++ wrapper for the intrinsics so that it throws, in which case this might be easier in the future anyways, but yeah, the code you have here ignores the fact that a vmread could fail.

bfvmm/include/intrinsics/intrinsics_intel_x64.h
```diff
@@ -79,8 +84,18 @@ class intrinsics_intel_x64 : public intrinsics_x64
...
+    {
+        uint64_t val = 0;
+        __vmread(field, &val);
+        return val;
+    }
+
     virtual bool vmlaunch()
     { return __vmlaunch(); }
+
+    bool tmp;
```
rianquinn: ?

bfvmm/include/intrinsics/intrinsics_intel_x64.h
```diff
@@ -79,8 +84,18 @@ class intrinsics_intel_x64 : public intrinsics_x64
...
+    volatile uint64_t rsp;
```
rianquinn: ?

bfvmm/include/vcpu/vcpu_manager.h
```diff
@@ -77,6 +86,14 @@ class vcpu_manager
     ///
     vcpu_manager_error::type stop(int64_t vcpuid);
 
+    /// Promote vCPU
+    ///
+    /// Promote vCPU to host CPU state
+    ///
+    /// @param vcpuid the vcpu to promote
+    /// @return On success the function never returns
+    vcpu_manager_error::type promote_vcpu(int64_t vcpuid);
```
rianquinn: missing extra ///

bfvmm/include/vmcs/bitmap.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <stdint.h>
+
+class bitmap
+{
+public:
+    /// Constructor
+    /// @param num_bits size of the bitmap in bits
+    ///
+    bitmap(uint32_t num_bits);
```
rianquinn: should be uint64_t. The only time we have been using 32bit types is if the data we are pointing to is 32bits.

bfvmm/include/vmcs/bitmap.h
```diff
@@ -0,0 +1,66 @@
...
+    /// address
+    /// @return the virtual address of the beginning
+    ///         of the bitmap
+    ///
+    uint8_t *address();
+
+    /// set_bit
+    /// @param n nth bit to set in the bitmap
+    ///
+    void set_bit(uint32_t n);
```
rianquinn: samething, everything should be uint64_t

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -58,6 +60,9 @@ class vmcs_intel_x64
     ///
     virtual vmcs_error::type launch();
 
+    virtual vmcs_error::type unlaunch();
```
rianquinn: why not call this promote?

bfvmm/src/vmcs/src/bitmap.cpp
```diff
@@ -0,0 +1,50 @@
...
+void bitmap::clear_bit(uint32_t n)
+{
+    if ((n / 8) > m_length) return;
+
+    m_bitmap[n / 8] &= ~(1 << (n % 8));
+}
+
+bool bitmap::bit(uint32_t n)
+{
+    if ((n / 8) > m_length) return false;
```
rianquinn: There are a couple of issues here. First, we should really try to avoid / in the hypervisor as it "could" cause problems. In this case it would not as you are using a constant.

Are you sure this actually works? Seems to me that if n == 2, you get 2/8 = 0, and if n == 3, you get 3/8 == 0.

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -294,6 +299,8 @@ class vmcs_intel_x64
 
     friend class vmcs_ut;
 
+    bitmap *m_msr_bitmap;
```
rianquinn: Why is the bitmap a pointer? At minimum, it should be a make_shared, but I would imagine this could be done without a pionter

bfvmm/src/exit_handler/src/Makefile
```diff
@@ -63,6 +63,7 @@ SOURCES+=exit_handler.cpp
 SOURCES+=exit_handler_entry.asm
 SOURCES+=exit_handler_dispatch.cpp
```
rianquinn: extra space

bfvmm/src/exit_handler/src/exit_handler.cpp
```diff
@@ -1,3 +1,4 @@
+
```
rianquinn: extra space

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
```diff
@@ -374,7 +365,24 @@ exit_handler_dispatch::handle_rsm()
...
+        case VMCS_PROMOTION:
+        {
+            advance_rip();
+            g_guest_rax = 0x00;
+            g_vcm->promote_vcpu(0);
+            break;
+        }
+        default:
+        {
+            g_guest_rax = 0xFFFFFFFFFFFFFFFF;
```
rianquinn: nice!!!

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
```diff
@@ -589,7 +597,8 @@ exit_handler_dispatch::handle_control_register_accesses()
     std::cout << std::dec << std::endl;
 
     spin_wait();
-    m_intrinsics_intel_x64->halt();
+    spin_wait();
```
rianquinn: why wait twice?

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
```diff
@@ -733,6 +748,20 @@ exit_handler_dispatch::unimplemented_handler()
 {
     std::cout << std::endl;
     std::cout << std::endl;
+    std::cout << "Guest register state: " << std::hex << std::endl;
+    std::cout << "----------------------------------------------------------------------" << std::endl;
+    std::cout << "g_guest_rax: 0x" << g_guest_rax << std::endl;
+    std::cout << "g_guest_rbx: 0x" << g_guest_rbx << std::endl;
+    std::cout << "g_guest_rcx: 0x" << g_guest_rcx << std::endl;
+    std::cout << "g_guest_rdx: 0x" << g_guest_rdx << std::endl << std::endl;
```
rianquinn: why endl twice?

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
```diff
@@ -741,7 +770,8 @@ exit_handler_dispatch::unimplemented_handler()
     std::cout << std::dec << std::endl;
 
     spin_wait();
-    m_intrinsics_intel_x64->halt();
+    spin_wait();
```
rianquinn: wait twice again

bfvmm/src/memory_manager/src/Makefile
```diff
@@ -60,6 +60,7 @@ CROSS_OUTDIR:=../../../bin
 ################################################################################
 
 SOURCES+=memory_manager.cpp
+HEADERS=
```
rianquinn: ?

bfvmm/src/memory_manager/src/memory_manager.cpp
```diff
@@ -246,6 +246,7 @@ memory_manager::virt_to_phys(void *virt)
 void *
 memory_manager::phys_to_virt(void *phys)
 {
+
```
rianquinn: extra spcae

bfvmm/src/memory_manager/src/memory_manager.cpp
```diff
@@ -305,6 +306,7 @@ memory_manager::add_mdl(struct memory_descriptor *mdl, int64_t num)
 
         m_virt_to_phys_map[(uintptr_t)md.virt >> MAX_PAGE_SHIFT] = md;
         m_phys_to_virt_map[(uintptr_t)md.phys >> MAX_PAGE_SHIFT] = md;
+
```
rianquinn: extra space

bfvmm/src/memory_manager/test/Makefile
```diff
@@ -51,6 +51,7 @@ NATIVE_OUTDIR:=../bin
 
 SOURCES+=test.cpp
 SOURCES+=test_memory_manager.cpp
+HEADERS=
```
rianquinn: ?

bfvmm/src/vcpu/src/vcpu_intel_x64.cpp
```diff
@@ -55,11 +56,20 @@ vcpu_intel_x64::vcpu_intel_x64(int64_t id,
 vcpu_error::type
 vcpu_intel_x64::start()
 {
-    // if (m_vmm.start() != vmm_error::success)
-    //     return vcpu_intel_x64_error::failure;
+    std::cout << "About to start and launch the hypervisor" << std::endl;
```
rianquinn: this extra cout should be removed. Text that we plan to keep around forever should also use the debug.h stuff instead of cout manually

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -164,20 +165,7 @@ vmcs_intel_x64::launch()
...
-    //     buf[i] = 0;
-
-    // vmwrite(VMCS_ADDRESS_OF_MSR_BITMAPS_FULL, (uint64_t)m_msr_bitmap.phys_addr());
-
-    // auto controls = vmread(VMCS_PRIMARY_PROCESSOR_BASED_VM_EXECUTION_CONTROLS);
-
-    // controls |= VM_EXEC_P_PROC_BASED_USE_MSR_BITMAPS;
-
-    // vmwrite(VMCS_PRIMARY_PROCESSOR_BASED_VM_EXECUTION_CONTROLS, controls);
+    vmwrite(VMCS_ADDRESS_OF_MSR_BITMAPS_FULL, (uint64_t)g_mm->virt_to_phys(m_msr_bitmap->address()));
```
rianquinn: This needs to be moved to the other VMCS functions were we set the fields that this belongs too

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -298,36 +285,51 @@ vmcs_intel_x64::save_state()
...
+    m_intrinsics->write_msr(IA32_PAT_MSR, m_intrinsics->vmread(VMCS_GUEST_IA32_PAT_FULL));
+    m_intrinsics->write_msr(IA32_SYSENTER_CS_MSR, m_intrinsics->vmread(VMCS_GUEST_IA32_SYSENTER_CS));
+    m_intrinsics->write_msr(IA32_FS_BASE_MSR, m_intrinsics->vmread(VMCS_GUEST_FS_BASE));
+    m_intrinsics->write_msr(IA32_GS_BASE_MSR, m_intrinsics->vmread(VMCS_GUEST_GS_BASE));
+    m_intrinsics->write_msr(IA32_SYSENTER_ESP_MSR, m_intrinsics->vmread(VMCS_GUEST_IA32_SYSENTER_ESP));
+    m_intrinsics->write_msr(IA32_SYSENTER_EIP_MSR, m_intrinsics->vmread(VMCS_GUEST_IA32_SYSENTER_EIP));
+    m_intrinsics->write_cr3(m_intrinsics->vmread(VMCS_GUEST_CR3));
+
+    promote_vmcs_to_root();
+
```
rianquinn: I would get rid of the extra space and comment. Instead, consider marking as noreturn

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -347,7 +349,7 @@ vmcs_intel_x64::create_vmcs_region()
 vmcs_error::type
 vmcs_intel_x64::release_vmxon_region()
 {
-    // m_memory_manager->free_page(m_vmcs_region);
+    delete m_vmcs_region;
```
rianquinn: please change to a make_unique so that we do not have to delete manually. I would like to stick to the "no naked new" rule if possible. In this case, you might be able to use std::array, but if that doesn't work, make_unique<char[]>(4096) will allocate a C style char array of the size your looking for.

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -355,19 +357,19 @@ vmcs_intel_x64::release_vmxon_region()
 vmcs_error::type
 vmcs_intel_x64::clear_vmcs_region()
 {
-    // auto phys = m_vmcs_region.phys_addr();
+    auto phys = memory_manager::instance()->virt_to_phys(m_vmcs_region);
```
rianquinn: so cool to see that working. :)

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -673,14 +675,15 @@ vmcs_intel_x64::write_natural_width_host_state_fields()
 {
     vmwrite(VMCS_HOST_CR0, m_cr0);
     vmwrite(VMCS_HOST_CR3, m_cr3);
+
```
rianquinn: extra space?

bfvmm/src/vmm/src/vmm_intel_x64.cpp
```diff
@@ -356,34 +357,25 @@ vmm_intel_x64::disable_vmx_operation()
...
-
-    // if (((uintptr_t)m_vmxon_page.phys_addr() & 0x0000000000000FFF) != 0)
-    // {
-    //     std::cout << "create_vmxon_region failed: "
-    //               << "the allocated page is not page aligned:" << std::endl
-    //               << "    - page phys: " << m_vmxon_page.phys_addr()
-    //               << std::endl;
-    //     return vmm_error::not_supported;
-    // }
+    m_vmxon_page = new char[4096];
```
rianquinn: same thing as above. Either use std::array if it will work, or make_unique<char[]>(4096)

bfvmm/src/vmm/src/vmm_intel_x64.cpp
```diff
@@ -356,34 +357,25 @@ vmm_intel_x64::disable_vmx_operation()
...
+    if (((uintptr_t)g_mm->virt_to_phys(m_vmxon_page) & 0x0000000000000FFF) != 0)
+    {
+        std::cout << "create_vmxon_region failed: "
+                  << "the allocated page is not page aligned:" << std::endl
+                  << "    - page phys: " << g_mm->virt_to_phys(m_vmxon_page)
+                  << std::endl;
+        return vmm_error::not_supported;
+    }
 
     // auto buf = (char *)m_vmxon_page.virt_addr();
```
rianquinn: why leave commented code here?

bfvmm/src/vmm/src/vmm_intel_x64.cpp
```diff
@@ -403,7 +395,7 @@ vmm_intel_x64::create_vmxon_region()
 vmm_error::type
 vmm_intel_x64::release_vmxon_region()
 {
-    // m_memory_manager->free_page(m_vmxon_page);
+    delete m_vmxon_page;
```
rianquinn: should be able to remove delete

bfvmm/src/vmm/src/vmm_intel_x64.cpp
```diff
@@ -411,27 +403,28 @@ vmm_intel_x64::release_vmxon_region()
...
 
-    // m_vmxon_enabled = true;
-    // std::cout << "vmxon: success" << std::endl;
+    m_vmxon_enabled = true;
+    std::cout << "vmxon: success" << std::endl;
 
     return vmm_error::success;
 }
 
+
```
rianquinn: extra space

driver_entry/src/arch/linux/entry.c
```diff
@@ -174,6 +174,7 @@ ioctl_stop_vmm(void)
     long status = BF_IOCTL_SUCCESS;
 
     ret = common_stop_vmm();
+
```
rianquinn: extra space

tools/scripts/create-cross-compiler.sh
```diff
@@ -7,7 +7,7 @@
 #     is mainly used by Travis CI because it cannot handle all of the output
 #     that this scripts spits out.
 #
-# export KEEP_TMP=true
+export KEEP_TMP=true
```
rianquinn: I would leave this turned off by default

.gitignore
brendank310: That should be an exception (and actually shouldn't happen at all). The contents of bin/ folders should be generated by the build and that's it.

README.md
brendank310: Alright makes sense.

bfvmm/include/intrinsics/intrinsics_intel_x64.h
brendank310: I think that's a 'buyer beware' version of the function in my opinion. If you use it, you are taking on the risk that the read could have failed. A similar function is provided in the VMCS class which sets a global bool flag (which is essentially ignored until all errors have coalesced together). Without a function like this, the code size balloons, with it there are runtime risks. Perhaps there should be a policy or main direction for this trade off in general.

bfvmm/include/intrinsics/intrinsics_intel_x64.h
brendank310: Cruft, will remove.

bfvmm/include/intrinsics/intrinsics_intel_x64.h
brendank310: Same as above.

bfvmm/include/vcpu/vcpu_manager.h
brendank310: Or maybe there is just an extra /// all over the project :D

bfvmm/include/vmcs/bitmap.h
brendank310: No, the num_bits controls an allocation, and there is no practical reason why we should be supporting bitmaps that have more than a couple billion bits in it.

bfvmm/include/vmcs/bitmap.h
brendank310: See above.

.gitignore
rianquinn: can't, we have makefiles in them for testing

bfvmm/include/vmcs/vmcs_intel_x64.h
brendank310: I wobbled back and forth on this, I'd be fine with promote (why the doc is missing here)

bfvmm/include/intrinsics/intrinsics_intel_x64.h
rianquinn: I don't think it really matters once the exception logic is complete. If the read fails, the exception logic will kill your attempt to continue unless you explicitly catch the failure, and manually change the corrupted bit in the VMCS.

bfvmm/src/vmcs/src/bitmap.cpp
brendank310: The division indexes the byte within which the bits lie. So 2 and 3 both lie in the first byte (byte 0). Note taken on using division, will explicitly use the shift operator instead so the compiler can't get stupid with power of 2 division.

bfvmm/include/vcpu/vcpu_manager.h
rianquinn: lol.... yeah that's standard practice for Doxygen with ///. Note sure why.

bfvmm/include/vmcs/vmcs_intel_x64.h
brendank310: I was getting bitched at by the compiler at one point about it's initialization as an object, but changing it now seems fine.

bfvmm/include/vmcs/bitmap.h
rianquinn: agreed there is no reason to alloc that much memory, but if that were your reason, you should change it to 16bits as that's more realistic. My main reason for enforcing all types to be 64bit is it's faster on 64bit, and I want to limit how much we are arbitrarily defining type sizes. Basically.... unless you specifically need it to be less than 64 bits (for example, the use of uint8_t * is fine), just use 64bits.

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
brendank310: At some point I think it would be good to have a class that handles vmcalls, so different modules could register hooks for different actions, but that's a discussion for another day.

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
brendank310: Easier on my eyes.

bfvmm/src/memory_manager/src/Makefile
brendank310: Cruft from me excising the paging code, and not trusting git after it destroyed my rebase yesterday afternoon.

bfvmm/src/exit_handler/src/exit_handler_dispatch.cpp
rianquinn: agreed 100%

.gitignore
brendank310: Being that they are already indexed, and boilerplate, it's not an issue. Having said that, I can add an exception to the gitignore for Makefiles

bfvmm/include/vmcs/bitmap.h
brendank310: 16bits actually only allows for a 2 page bitmap, and for example, (the EOI exit bitmap appears to be 4 pages)

PR #62, opened 2016-02-28
Exception Support

rianquinn:
The following patches provide bareflank with exception support.
This includes:

- A custom unwinder that is capable of unwinding the stack with
  thread-saftey, while only requiring libc (which can also be
  disabled)
- A set of unit tests to validate that throw exceptions works
- Changes to the VMM to support the custom unwinder, and
  install catch all blocks incase an exception is not
  caught.

### DIFF
--- a/bfunwind/src/dwarf4.cpp
+++ b/bfunwind/src/dwarf4.cpp
@@ -1,0 +1,575 @@
+//
+// Bareflank Unwind Library
+//
+// Copyright (C) 2015 Assured Information Security, Inc.
+// Author: Rian Quinn        <quinnr@ainfosec.com>
+// Author: Brendan Kerrigan  <kerriganb@ainfosec.com>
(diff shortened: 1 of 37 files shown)


bfunwind/src/dwarf4.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+    return row;
+}
+
+// -----------------------------------------------------------------------------
+// DWARF 4 Implementation
+// -----------------------------------------------------------------------------
+
+int64_t
+dwarf4::dwarf4::decode_sleb128(char **addr)
```
brendank310: Double classname?

rianquinn: yeah I fixed that in the PR that follows. I have no idea how that compiles, but yeah, it has been fixed.

PR #70, opened 2016-03-04
VMCS Exception Support

rianquinn:
This patch provides VMCS exception support. During it's
development, it was also identified that the libc++ code can
overwrite the kernel's stack as it's expecting a much larger
stack than the Linux kernel is providing. To overcome this
issue, this patch also includes code to guard the kernel's
by using a different stack while the VMM is execuitng.


brendank310:
LGTM

PR #72, opened 2016-03-05
VMCS Supported Checks

rianquinn:
This patch provides checks for which CPU features are supported.
Future patches will leverge these checks to prevent tests from
being run when the CPU doesn't support the feature


brendank310:
A little verbose on the supports_blah stuff, but over all looks good to me.

PR #75, opened 2016-03-12
Vmcs guest checks

rianquinn:


bfvmm/src/vmxon/src/vmxon_intel_x64.cpp
```diff
@@ -172,14 +168,15 @@ vmxon_intel_x64::create_vmxon_region()
     auto cor1 = commit_or_rollback([&]
     { this->release_vmxon_region(); });
 
-    m_vmxon_region = std::make_unique<char[]>(4096);
-    m_vmxon_region_phys = (uintptr_t)g_mm->virt_to_phys(m_vmxon_region.get());
+    auto region = (uint32_t *)g_mm->malloc_aligned(4096, 4096);
```
brendank310: I think if we're going to be explicit about page aligned allocations, we should make that the standard practice throughout the project. Anything page aligned should follow the above pattern (malloc_aligned and then wrap in appropriate ptr.)

brendank310:
LGTM

bfvmm/src/vmxon/src/vmxon_intel_x64.cpp
rianquinn: agreed

PR #81, opened 2016-03-21
Vcpu manager unittests

rianquinn:


brendank310:
Was merged as part of a different pull request (weird commit hash problem).

PR #84, opened 2016-03-23
Store and restore guest PAT MSR values

brendank310:
If you don't store and restore guest PAT MSR values, system performance after turning off the hypervisor is pretty awful, as it seems to default to all no caching.


rianquinn:
+1

PR #86, opened 2016-03-24
Remove src

rianquinn:


brendank310:
Was merged as part of a different pull requests (weird commit hash problem).

PR #89, opened 2016-03-24
[vmcs/vmxon] Clean up the VMX TRUE controls to use the correct naming…

brendank310:
…, and test for the capability in vmxon


rianquinn:
+1

PR #91, opened 2016-03-26
Div mod cleanup

rianquinn:


brendank310:
We may want to consider alignment always being expressed as a power of 2, which could avoid some of the checking that is necessary (e.g. 12 instead of 4096). This could be done at a later time and cleanup however. This will be closed, but merged as part of the rebased PR #95.

PR #92, opened 2016-03-26
Invd instruction

rianquinn:


brendank310:
LGTM. Same deal with the previous two commits though, closing and will merge as part of #95.

Issue #98, opened 2016-04-03
Linux driver compilation fails on older kernels


rianquinn:
The Linux driver does not compile on older kernels because cr4_init_shadow does not exist. It appears that this optimization was added in Linux 4.0

rianquinn:
Has been fixed

PR #99, opened 2016-04-03
Fix Driver Compile Bug

rianquinn:
This patch fixes an issue with the Linux kernel that prevented
the Bareflank driver from compiling on older versions of Linux
as the cr4_init_shadow did not exist until Linux 4.0

[issue]: https://github.com/Bareflank/hypervisor/issues/98


rianquinn:
All checks pass, merging.

PR #109, opened 2016-04-23
Fix CTORS/DTORS

rianquinn:
The local_init and local_fini functions were being executed out of
the info struct which (besides from being an ugly approach) was
causing issues with no-execute stacks. There was also a bug where
destructors were not being executed which was also fixed.


brendank310:
LGTM

PR #114, opened 2016-05-09
Osx support

brendank310:
Experimental support for OSX. Currently cannot successfully load modules, but bfm to driver are passing data to each other.


Makefile
```diff
@@ -64,6 +64,9 @@ include ./common/common_subdir.mk
 .PHONY: linux_load
 .PHONY: linux_unload
 .PHONY: linux_clean
+.PHONY: osx_load
+.PHONY: osx_unload
+.PHONY: osx_clean
```
rianquinn: please remove these. these will be removed for linux as well, and will be changed to driver_load, driver_unload at some point with the goal that is the makefile can do it, it will detect the OS and run the proper commands for you

Makefile
```diff
@@ -92,6 +95,16 @@ linux_unload: force
 
 linux_clean: linux_unload
 
+osx_load: force
```
rianquinn: same as above

bfdrivers/src/Makefile
```diff
@@ -78,6 +78,12 @@ LINUX_INCLUDE_PATHS+=
 LINUX_LIBS+=
 LINUX_LIBRARY_PATHS+=
 
+OSX_SOURCES+=
```
rianquinn: please remove, I will be adding these at some point, but this will mess with the merge

bfdrivers/src/arch/osx/fixup-kext-perms.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+sudo chown -R root /Users/brendank310/Library/Developer/Xcode/DerivedData/bareflank-*/Build/Products/Debug/bareflank.kext
```
rianquinn: is this really needed?

common/common_target.mk
```diff
@@ -50,6 +50,8 @@ else
 	UNAME:=$(shell uname -s)
 	ifeq ($(UNAME), Linux)
 		OS:=linux
+	else
```
rianquinn: should be: ifeq ($(UNAME), Darwin)

common/common_target.mk
```diff
@@ -138,7 +140,7 @@ CROSS_ASMFLAGS+=-f elf64
 # Default AR Flags
 ################################################################################
 
-NATIVE_ARFLAGS+=rcs
+NATIVE_ARFLAGS+=-rcs
```
rianquinn: did we test to make sure this works on Linux?

common/common_target.mk
```diff
@@ -254,6 +272,11 @@ ifeq ($(TARGET_NATIVE_COMPILED), true)
 		NATIVE_LIBRARY_PATHS+=$(LIBRARY_PATHS)
 		NATIVE_LIBRARY_PATHS+=$(LINUX_LIBRARY_PATHS)
 	endif
+	ifeq ($(OS), osx)
+		NATIVE_LIBRARY_PATHS+=$(LIBRARY_PATHS)
+		NATIVE_LIBRARY_PATHS+=$(OSX_LIBRARY_PATHS)
+		NATIVE_LDFLAGS+=$(OSX_LDFLAGS)
```
rianquinn: we need a better place to put this. commit this for now, and I will add a section for the cc, cxx and ld flags to be OS specific as well.

include/driver_entry_interface.h
```diff
@@ -167,6 +176,24 @@ extern "C" {
...
+#include <stdint.h>
+
+typedef struct bf_ioctl
+{
+    uint32_t command;
+    uint32_t size;
+    void *addr;
+} bf_ioctl_t;
+
+#define IOCTL_ADD_MODULE_LENGTH IOCTL_ADD_MODULE_LENGTH_CMD
```
rianquinn: Before we complete 1.1, we should document these, even if the documentation is the same as Linux

bfdrivers/src/arch/osx/fixup-kext-perms.sh
brendank310: I'd argue yes, though I think I may have updated the unfixup-kext-perms and not this to remove the hardcoded path. Without it, you cannot load up the kext. You would think XCode would change the perms after a build, but it currently does not. I don't know a lot about XCode but perhaps the project itself can have a pre/post build hook to change the permissions.

common/common_target.mk
brendank310: Just ran this branch through the gamut of tests on the Windows side, seems fine.

common/common_target.mk
brendank310: I had originally considered that, but I think this section needs to be reworked over all (essentially making the Windows case fall in line with the pattern as well).

rianquinn: agreed. The Windows side is really Cygwin / Babun, so it needs some work, but yeah, uname exists on OSX

common/common_target.mk
brendank310: Make that the Linux side (working on Windows stuff at the same time 👎 )

PR #116, opened 2016-06-20
New Build System

rianquinn:
This patch provides a new build system that:
- provides support for docker to decrease the time it takes to
  run a build. a local compiler can still be used if desired
- provide better support for other platforms
- provides support for out-of-tree compilation which also
  allows for multiple build trees if desired


tools/dockerfiles/dockerfile_gcc_540
```diff
@@ -0,0 +1,82 @@
...
+
+COPY ./tools/scripts/build_binutils.sh /tools/scripts/build_binutils.sh
+COPY ./tools/scripts/build_gcc.sh /tools/scripts/build_gcc.sh
+COPY ./tools/scripts/build_nasm.sh /tools/scripts/build_nasm.sh
+
+# ------------------------------------------------------------------------------
+# Script
+# ------------------------------------------------------------------------------
+
+RUN CROSS_COMPILER=gcc_540 ./tools/scripts/create-cross-compiler.sh
```
ranok: Usually you add a VOLUME and WORKDIR to the Dockerfiles to ensure there is a partition there like /bareflank and it will automatically run in that directory if you add: 
VOLUME /bareflank
WORKDIR /bareflank

This reduces the worry that when you use the docker file in a script a docker update might change the default working dir if  you mount stuff into the container.

rianquinn: At the moment we do this on the command line with -v, and the workdir is /. The dockerfile is only used to create the images that are placed in docker hub, which will only be run by the maintainers.

brendank310:
Merged as part of the isolation pull requests.

PR #134, opened 2016-07-23
DWARF Expression Suport

rianquinn:
GCC 6.1 appears to add DWARF expressions for stack realignment
which is needed to support the addition of SSE instructions.
This patch provides the needed expression support for exceptions.
It should be noted that there appears to be a bug in GCC 6.1
that we have hacked around for now. For more information, please
see:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71978


bfunwind/src/dwarf4.cpp
```diff
@@ -183,22 +202,1287 @@ class cfi_table_row
...
+//       class that guards [] to prevent the stack variable from being
+//       accessed out of bounds
+
+static uint64_t
+private_parse_expression(char *p,
+                         uint64_t initialStackValue,
+                         register_state *state)
+{
+    uint64_t i = 0;
+    uint64_t stack[100];
```
brendank310: This is a strange number, why 100?

bfunwind/src/dwarf4.cpp
```diff
@@ -30,14 +30,22 @@
...
         b \
         return; \
     }
 
+#define if_opcode(a,b) \
+    if (opcode == a) \
+    { \
+        log("    - %s: ", #a); \
+        b \
+        continue; \
```
brendank310: Why is this continue but if_cfa is return? Never a huge fan of hiding a branch or jump in a macro...

bfunwind/src/dwarf4.cpp
rianquinn: Please see: https://github.com/llvm-mirror/libunwind/blob/master/src/DwarfInstructions.hpp#L228

bfunwind/src/dwarf4.cpp
rianquinn: if_opcode is in a while loop. The goal of the macro is to simplify the amount of cut / past there is. Looking at Apple's implementation, there are so many CFA and Expression opcodes, writing "continue" / "return" every time distracts from the point of the logic, but yeah, normally I would agree. For an example, please see: https://github.com/llvm-mirror/libunwind/blob/master/src/DwarfParser.hpp#L435

bfunwind/src/dwarf4.cpp
brendank310: I see that they use 100, but the why is still alluding me.

rianquinn: The DWARF spec does not define a stack size. You have to pick what you want to support. We just do the same thing Apple is doing

brendank310: Like I said, that's fine, but it should still be something along the lines of:
#define STACK_SIZE 100
Not only does that make it easy to change in the future if necessary, but it provides a little clearer picture as to what its source is (arbitrary size in this instance.)

rianquinn:
I added "#define EXPRESSION_STACK_SIZE 100" and rebased

bfunwind/src/dwarf4.cpp
rianquinn: fixed

PR #137, opened 2016-07-26
Posix Thread Mutex Support

rianquinn:
This pull request adds pthread support, which enables the use
of std::mutex and friends. Only the needed functions are used
so attempts to use std::thread will fail as this only adds
mutex support.

### DIFF
--- a/KNOWN_ISSUES.md
+++ b/KNOWN_ISSUES.md
@@ -5,14 +5,6 @@
   disable this check from executing or don't use a kernrel with this enabled.
   This is seen on Fedora as installing the kernel source enables a debug kernel
   by default with this enabled.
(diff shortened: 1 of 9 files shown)


brendank310:
@rianquinn this needs a rebase, and nice job sneaking the actual guts of the change in on libbfc 😕

rianquinn:
Hahaha…. yeah sorry about the bfc merge. I don’t have a good way of testing that with my own repo yet, and it will not be a problem in the future once we have our own libc and get rid of newlib as everything will reside in our repo. At some point, we should consider a config file that is not in the repo that defines things like C macros and repos for this very reason.

rianquinn:
rebased

PR #138, opened 2016-07-26
initial support for openSUSE

spockfish:
This adds initial support for openSUSE tumbleweed. 

More work needs to happen, as I think in general the dependancy of the LSB stuff should be phased out. As the supported distro's are all systemd based I think we could rely on '/etc/os-release' in stead.

### DIFF
--- a/tools/scripts/setup_opensuse.sh
+++ b/tools/scripts/setup_opensuse.sh
@@ -1,0 +1,172 @@
+#!/bin/bash -e
+#
+# Bareflank Hypervisor
+#
+# Copyright (C) 2015 Assured Information Security, Inc.
+# Author: Rian Quinn        <quinnr@ainfosec.com>
+# Author: Brendan Kerrigan  <kerriganb@ainfosec.com>
+#
+# This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


brendank310:
If os-release is where things are going (hard to keep up to date with all the nooks and crannies that are changing as the move to systemd matures) then I definitely agree. Perhaps the setup-suse script will be a good quick test of this move, then we can bring the change to the other scripts? One thing we have to consider is the TravisCI baseline that is used for testing. Maybe @rianquinn can comment on that?

brendank310:
@spockfish I'm going to merge this in, would you be willing to field questions on suse if they come up?

rianquinn:
TravisCI uses Ubuntu 14.04 since we use Docker to reduce the time it takes to get the dev environment up and running. I'm not sure if os-release is available on 14.04 so we will have to look into that. So long as it is available, I am a huge fan of getting rid of old stuff.

spockfish:
ok cool. I'll have a look on SUSE and check if Ubuntu 14.04 has this already in place. Maybe this could end up with one linux script to rule them all ;-)

PR #144, opened 2016-07-27
Error Code Cleanup

rianquinn:
Originally, the error codes were self contained which made sense
at the time, but as the project has grown, the number of error
codes has increased leading to a lot of redudancy in common.c as
well as overlapping error codes in a couple of cases. This patch
moves all of the error codes toa single file, makes sure the
return types are consistent, and provides a single error code
to string conversion function

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -19,9 +19,7 @@ Misc:
   flags for BFM and windows for export all.
 
 Version 1.1 TODO:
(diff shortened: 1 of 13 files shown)


brendank310:
@rianquinn Please rebase.

rianquinn:
@brendank310 rebased

PR #152, opened 2016-07-29
Cleanup Windows IOCTL Interface

rianquinn:
This patch cleans up the Windows IOCTL interface. One of the
IOCTLs is not used as it was Linux specific, and the others
were hardcoded, and some had too many permissions.

### DIFF
--- a/include/driver_entry_interface.h
+++ b/include/driver_entry_interface.h
@@ -56,18 +56,18 @@ extern "C" {
 #endif
 
 #ifndef BAREFLANK_DEVICETYPE
-#define BAREFLANK_DEVICETYPE 0xf00d
+#define BAREFLANK_DEVICETYPE 0xF00D
 #endif
 
-#define IOCTL_ADD_MODULE_LENGTH_CMD 101
-#define IOCTL_ADD_MODULE_CMD 100
-#define IOCTL_LOAD_VMM_CMD 200
-#define IOCTL_UNLOAD_VMM_CMD 300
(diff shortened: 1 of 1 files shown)


include/driver_entry_interface.h
```diff
@@ -210,7 +200,7 @@ DEFINE_GUID(GUID_DEVINTERFACE_bareflank,
  * @param arg character buffer containing the module to add
  * @return
  */
-#define IOCTL_ADD_MODULE CTL_CODE(BAREFLANK_DEVICETYPE, 0x901, METHOD_IN_DIRECT, FILE_WRITE_DATA)
+#define IOCTL_ADD_MODULE CTL_CODE(BAREFLANK_DEVICETYPE, IOCTL_ADD_MODULE_CMD, METHOD_IN_DIRECT, FILE_WRITE_DATA)
```
brendank310: These values are reserved for Microsoft. 

>FunctionCode

Identifies the function to be performed by the driver. Values of less than 0x800 are reserved for Microsoft. Values of 0x800 and higher can be used by vendors. Note that the vendor-assigned values set the Custom bit.https://msdn.microsoft.com/en-us/library/windows/hardware/ff543023(v=vs.85).aspx

rianquinn: fixed

rianquinn:
fixed IOCTL defines to support windows, rebased to the latest, and tested to make sure they work.

PR #155, opened 2016-07-31
Add JSON Support

rianquinn:
To support generic configuration files, and commands to the
hypervisor, we will support JSON. A generic VMCall will also
be supported, but unlikely to be used based on how easy
JSON is.

To support JSON, we are using a modified version of:
https://github.com/nlohmann/json

This is a really well tested JSON, header only implementation.
In addition, this implementation already supports native
integer types. Gutting float / double from this implementation
was trivial, meaning it will be easy to support newer versions
in the future.

To verify this works, double / float was gutted, and all of the
tests on the webstie were tested inside the VMM, and they all
worked great. If you attempt to use a floating point, the
compiler will detect this and error out.

Future PRs will add support for start_vmm, create_vcpu,
make_vcpu, and vmcall to support JSON, allowing extensions
to customize how this logic works. We will also provide
additional interfaces on BFM and the drivers to support
this.

### DIFF
--- a/include/json.h
+++ b/include/json.h
@@ -1,0 +1,10035 @@
+/*
+    __ _____ _____ _____
+ __|  |   __|     |   | |  JSON for Modern C++
+|  |  |__   |  |  | | | |  version 2.0.2
+|_____|_____|_____|_|___|  https://github.com/nlohmann/json
+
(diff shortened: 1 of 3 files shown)


rianquinn:
Moving this code to the extended APIs so that JSON is optional

PR #165, opened 2016-08-31
Test vmlaunch failure

connojd:
This patch tests a vmlaunch failure and ensures all checks pass in the vmcs launch function.  Also, tests were added to reach full coverage of vmcs_intel_x64.cpp


bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -43,6 +80,16 @@ write_msr(uint32_t msr, uint64_t val)
 static bool
 vmread(uint64_t field, uint64_t *val)
 {
+    if (VMCS_VM_INSTRUCTION_ERROR == field)
+    {
+        vmcs_vm_instruction_error_read_count++;
+
+        if (vmcs_vm_instruction_error_read_count == 2)
```
rianquinn: This is really over-specified.... would love to see an alternative to this. Specifically, if the implementation changes, this unit test will fail even if the code is still valid.

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -193,22 +448,38 @@ setup_vmcs_x64_state_intrinsics(MockRepository &mocks, vmcs_intel_x64_state *sta
 }
 
 static void
+setup_vmcs_launch_failure(MockRepository &mocks, intrinsics_intel_x64 *in)
+{
+    setup_msrs();
+    setup_vmcs_fields();
+
+    Call &vmlaunch = mocks.OnCall(in, intrinsics_intel_x64::vmlaunch).Return(false);
+    mocks.OnCall(in, intrinsics_intel_x64::read_msr).After(vmlaunch).Do(read_msr);
```
rianquinn: since your calling read_msr, which was previously setup, what exactly is the reason for this code? I would prefer to see a flag in the read_msr code that you set saying something like `return_crap=true`

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -231,3 +502,308 @@ vmcs_ut::test_launch_success()
...
+    auto mm = mocks.Mock<memory_manager>();
+    auto in = bfn::mock_shared<intrinsics_intel_x64>(mocks);
+    auto host_state = bfn::mock_shared<vmcs_intel_x64_state>(mocks);
+    auto guest_state = bfn::mock_shared<vmcs_intel_x64_state>(mocks);
+
+    setup_vmcs_intrinsics(mocks, mm, in.get());
+    setup_vmcs_x64_state_intrinsics(mocks, host_state.get());
+    setup_vmcs_x64_state_intrinsics(mocks, guest_state.get());
+
+    auto fa = gsl::finally([&]
```
rianquinn: 😄

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -231,3 +502,308 @@ vmcs_ut::test_launch_success()
...
+vmcs_ut::test_launch_create_exit_handler_stack_failure()
+{
+    MockRepository mocks;
+    auto mm = mocks.Mock<memory_manager>();
+    auto in = bfn::mock_shared<intrinsics_intel_x64>(mocks);
+    auto host_state = bfn::mock_shared<vmcs_intel_x64_state>(mocks);
+    auto guest_state = bfn::mock_shared<vmcs_intel_x64_state>(mocks);
+
+    setup_vmcs_intrinsics(mocks, mm, in.get());
+    mocks.OnCallFunc(std::make_unique<char[]>).With(STACK_SIZE).Throw(std::exception());
```
rianquinn: wow... that works huh.... going to remember that

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -231,3 +502,308 @@ vmcs_ut::test_launch_success()
...
+    RUN_UNITTEST_WITH_MOCKS(mocks, [&]
+    {
+        vmcs_intel_x64 vmcs(in);
+
+        EXPECT_EXCEPTION(vmcs.launch(host_state, guest_state), std::logic_error);
+    });
+}
+
+void
+vmcs_ut::test_launch_check_host_state_failure()
```
rianquinn: the check_xxx_state_failure functions can be removed. Your going to do this for every single check, so it's assumed that it's working. That being said... keep the code you have because your testing at least one of them in each of these.

PR #179, opened 2016-09-05
Bad Alloc Testing

rianquinn:
This patch provides some small updates to enable catching and
handling when memory runs out (std::bad_alloc). Specifically,
when this happens, we can no longer use std::cout which
does many allocations in the process. Instead we exit
gracefully with an error code.


rianquinn:
Will post again once I can sort out the issues with the build. Ubuntu has updated the toolchain PPA and broke something (binutils and GCC do not seem to match at the moment as they are passing around unrecognized arguments to each other). My guess is future PRs will have this issue until we can resolve it.

Issue #176, opened 2016-09-03
Remove unnecessary null pointer checks


elfring:
[Extra null pointer checks are not needed](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for NULL before delete p?") in [the function "driver_entry_ut::fini"](https://github.com/Bareflank/hypervisor/blob/19f91d615a20b8c537383235d7634ad004126575/bfdrivers/test/test.cpp#L117 "Update candidate: driver_entry_ut::fini()").

rianquinn:
Agreed these should be removed. I'm actually surprised that it's not being caught by Clang Tidy. Currently we use 3.8, so it will be interesting to see if this sort of issue is caught in 3.9 (it appears that added a lot more [checks](http://llvm.org/releases/3.9.0/tools/clang/tools/extra/docs/ReleaseNotes.html) in 3.9)

rianquinn:
Clang Tidy readability states that nullptr checks should in fact be used regardless of the iso reference. Furthermore, the C++ Core Guidelines use this same syntax as well (although there doesn't seem to be a specific rule yet).

elfring:
* To which check of "[Clang Tidy](http://clang.llvm.org/extra/clang-tidy/checks/list.html "Clang-Tidy Checks")" would you like to refer here?
* Why do you prefer to [keep unnecessary null pointer checks before using the delete operator](https://github.com/rianquinn/hypervisor/commit/ff7882dbc9b6432268f64eef6730150b9bb15aa6#diff-75819e083e904a520ab8324af9ea8e82L118 "Update candidate: driver_entry_ut::fini()")?

RicoAntonioFelix:
@rianquinn I think that @elfring is proposing you just `delete[] entity` as doing `delete` on a `nullptr` is a `no-op`...

rianquinn:
My bad. I assumed that the issue was for 

```
if (blah == nullptr) -> if (blah)
```

And not the nullptr check itself. [readability-implicit-bool-cast] is the check I was referring to. I will provide a new PR that just removes the checks for `delete` and `delete[]`. For reference, CPP does in fact state that the nullptr check is redundant. 

http://en.cppreference.com/w/cpp/memory/new/operator_delete

rianquinn:
[...]

PR #189, opened 2016-09-07
Remove Redundant Null Pointer Checks

rianquinn:
Remove uses of == nullptr and != nullptr

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/176


rianquinn:
The readability checks are failing as they want explicit == and != nullptr. Furthermore, the C++ Core Guidelines use this same syntax.

Issue #185, opened 2016-09-06
Incorrect event injection instruction length check


connojd:
From section 26.2.1.3 from the [Intel manual](http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-manual-325462.pdf):

> If the interruption type is software interrupt, software exception, or privileged software exception, the
VM-entry instruction-length field is in the range 0–15. A VM-entry instruction length of 0 is allowed only if IA32_VMX_MISC[30] is read as 1; see Appendix A.6.

Should we read the MISC msr to make sure that a length of 0 isn't allowed?  Currently `check_control_event_injection_instr_length_checks` throws a `logic_error` if the length is 0. I noticed that IA32_VMX_MISC_MSR isn't in intrinsics_intel_x64.h so it would need to be added.

rianquinn:
Yeah, we can add that. I would put that in your PR for the control checks.

connojd:
Ok I was going to include fixes for each issue I find in the control checks in the PR

rianquinn:
Awesome 😄

rianquinn:
@connojd please close this once the unit test is complete and merged

rianquinn:
Fixed here: https://github.com/Bareflank/hypervisor/pull/209

Issue #183, opened 2016-09-06
Missing Page Modification Log check


connojd:
From section 26.2.1.1 of the [Intel manual](http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-manual-325462.pdf):

> If the "enable PML" VM-execution control is 1, the "enable EPT" VM-execution control must be 1.

This check, along with the check that the PML address is a 4KB-aligned physical address, is absent in vmcs_intel_x64_check_controls.cpp

rianquinn:
Yeah, that's a bug. They should be added.

rianquinn:
@connojd please close this once the unit test is complete and merged

rianquinn:
Fixed here: https://github.com/Bareflank/hypervisor/pull/209

Issue #170, opened 2016-09-02
[Question] Physical-address width in check_control_enable_ept_checks


connojd:
The check 
```cpp
if ((eptp & 0xFFFF000000000000) != 0 || (eptp & 0x0000000000000F80) != 0)
       throw std::logic_error("bits 11:7 and 63:48 of the eptp must be 0");
```
in `vmcs_intel_x64::check_control_enable_ept_checks` essentially hard-codes the physical address width.  Would it be better to call `is_physical_address_valid(eptp)` instead?

rianquinn:
Checking to make sure that the physical address is valid would take care of the upper bits, but it would not take care of the lower bits 11:7. You could break this up into two checks so that there isn't duplication with the physical address width.

rianquinn:
@connojd please close this once the unit test is complete and merged

rianquinn:
Fixed here: https://github.com/Bareflank/hypervisor/pull/209

Issue #168, opened 2016-09-02
[Question] Incorrect check for VM-execution control "use TPR shadow"


connojd:
Section 26.2.1.1, Volume 3C of the [Intel manual](http://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-manual-325462.pdf) states in the eighth bullet down,
> If the "use TPR shadow" VM-execution control is 1 and the "virtual-interrupt delivery" VM-execution control is 0, bits 31:4 of the TPR threshold VM-execution control field must be 0.

The next bullet states

> The following check is performed if the "use TPR shadow" VM-execution control is 1 and the "virtualize APIC accesses" and "virtual-interrupt delivery" VM-execution controls are both 0: the value of bits 3:0 of the TPR threshold VM-execution control field should not be greater than the value of bits 7:4 of VTPR (see Section 29.1.1).

The function that implements [the check](https://github.com/Bareflank/hypervisor/blob/master/bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp) is vmcs_intel_x64::check_control_tpr_shadow_and_virtual_apic.  It currently checks bits 31:4 of the TPR threshold only if TPR shadow is 1 and virtual-interrupt delivery is 1.  Shouldn't the check occur when TPR shadow is 1 and virtual-interrupt delivery is 0?

Likewise bits 3:0 of TPR threshold are checked against bits 7:4 of VTPR only if TPR shadow, virtual-interrupt delivery and virtualize APIC accesses are all 1 when the manual states the latter two should be 0.

rianquinn:
Yup... the code we have there is wrong.

rianquinn:
@connojd please close this once the unit test is complete and merged

rianquinn:
Fixed here: https://github.com/Bareflank/hypervisor/pull/209

PR #191, opened 2016-09-07
Fix vmcs check control issues

connojd:
This patch fixes the following issues found with the
vmcs control checks:

    Issue #168
    Issue #170
    Issue #183
    Issue #185


rianquinn:
LGTM +1

PR #202, opened 2016-09-10
view_as_pointer.h: Constrain utility function to integral types

RicoAntonioFelix:
Signed-off-by: “Rico <“ricoantoniofelix@yahoo.com”>

### DIFF
--- a/include/view_as_pointer.h
+++ b/include/view_as_pointer.h
@@ -22,6 +22,8 @@
 #ifndef VIEW_AS_POINTER
 #define VIEW_AS_POINTER
 
+#include <type_traits>
+
 // There are basically two different ways to print an integer as a pointer
 // (i.e. 0xXXXXXXXXXXXXXXXX).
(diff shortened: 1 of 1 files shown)


include/view_as_pointer.h
```diff
@@ -35,8 +37,8 @@
 //   dereferenced, it should not be an issue.
 //
 
-template<class T>
-void *view_as_pointer(const T &val)
+template<class T, class = typename std::enable_if<std::is_integral<T>::value>::type>
```
rianquinn: Nice catch!!!

That's pretty cool. Going to have to remember that. Here is a reference if anyone wants to know how this black magic works. 

https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/SFINAE

Now all that being said, I really wish they would have just added this to the language so the syntax was easier to understand. These sort of tricks work, but they are confusing.

PR #209, opened 2016-09-13
Add vmcs check control tests

connojd:
This patch tests every function defined in
vmcs_intel_x64_check_controls.cpp.  The way in which the tests
are implemented is different from other modules.

Each function-under-test's control flow graph is represented
by a vector of lambdas that modify or enable/disable VMCS fields
and MSRs. Each element in the function's cfg represents exactly one
path from its control flow graph.  Each function-under-test corresponds
to exactly one member of vmcs_ut.  This member enumerates over the
cfg of the function-under-test, testing each element (i.e. path) in turn
with RUN_UNITTEST_WITH_MOCKS.

There are several advantages to testing this way.  First, all of the
logic to test a function f is contained in its corresponding
setup_f_paths function, rather than being spread out across several
setup functions that setup vmcs fields for many functions unrelated to
the one under test.  This simplifies verification and debugging of the
test itself because you only have to look at one function to see how it
works.

Also, as the underlying source changes, the test for that source need
only change in one place.  Implementors of tests don't have to worry
about the flip of a bit in one VMCS field breaking other unit tests
for unrelated functions.  The only data dependencies that need to be
tracked down in the event of a change are right there in one setup
function.  This aids comprehensibility and maintainability of the tests.


connojd:
Astyle messed with the RUN_VMCS_TEST formatting a little bit too much.

bfvmm/src/vmcs/test/test_vmcs_intel_x64_check_controls.cpp
```diff
@@ -0,0 +1,1089 @@
...
+extern std::map<uint64_t, uint64_t> g_vmcs_fields;
+extern bool g_phys_to_virt_return_nullptr;
+
+bool vmread(uint64_t field, uint64_t *val);
+uint64_t read_msr(uint32_t msr);
+uint32_t cpuid_eax(uint32_t val);
+void *physint_to_virtptr(uintptr_t phys);
+
+#define RUN_VMCS_TEST(func) \
+    do \
```
rianquinn: Please convert this to a function. You could either use a template function, or if the signature is known, you could define the signature up front. That should clear up the Astyle issue as well. If it doesn't we can add blockers to Astyle as a last ditch effort.

bfvmm/src/vmcs/test/test_vmcs_intel_x64_check_controls.cpp
```diff
@@ -0,0 +1,1089 @@
...
+setup_check_control_proc_based_ctls2_reserved_properly_set_paths(std::vector<std::function<enum path_type()>> &cfg)
+{
+    cfg.push_back([&]
+    {
+        g_msrs[IA32_VMX_PROCBASED_CTLS2_MSR] = 0;
+        g_vmcs_fields[VMCS_SECONDARY_PROCESSOR_BASED_VM_EXECUTION_CONTROLS] = 0;
+        return TEST_SUCCESS;
+    });
+
+    cfg.push_back([&]
```
rianquinn: I like these!!!! 😄

rianquinn:
I like the approach here. It's clean and easy to read. I think the only mod that needs to be made is to cleanup the #define and you should be good to go.

connojd:
So as it stands right now, all of the check_* functions of vmcs have the same signature.  Do you think I should make a non-generic function for the test macro?  We could always write a generic one in the future if we wanted to used the method for functions with a different signature (such as vmcs::launch).

rianquinn:
Let's make it generic

connojd:
The run_vmcs_test has std::logic_error hardcoded as the exception type right now.  I think the EXPECT_EXCEPTION may have to be turned into a function to allow different exception types there. Any thoughts on the best avenue to allow different exception types?

rianquinn:
My advice would be to find a way to write these macros as functions, even if that means we have to find/replace a lot of text in the code to clean it up. The macros are my "C" habits showing, so yeah, we should get rid of them in favor of something that's more flexible and inline with C++.

connojd:
Is there a way to print __LINE__ info in a function? The if(!caught)....else... after the try-catch block in EXPECT_EXCEPTION uses __LINE__ and __PRETTY_FUNCTION__.
Would it be ok to make that if-else block a macro that is invoked after the function version of expect_exception?

rianquinn:
This is one area where languages like Swift have a leg up. Default arguments like \_\_LINE\_\_ don't work in C++ which sucks. The only solution I know of, is to use a macro. 

Now... that being said, I think its fair to use lower snake case names for macros that call a function that only exist to deal with \_\_LINE\_\_. So if you convert this to a function, and then create a macro function to provide \_\_LINE\_\_, I would still use lower case as it's really a function. The only thing the macro is doing is providing the default argument in way C++ can handle.

connojd:
I've updated the PR to include the generic functions unittest::expect_exception and unittest::expect_no_exception.  Currently they're only used in the vmcs control check tests.

include/unittest.h
```diff
@@ -522,15 +546,80 @@ class unittest
...
+                std::cout << "caught: " << e << '\n';
+        }
+        catch (...)
+        {
+            caught = true;
+            wrong_exception = true;
+            std::cerr << "unknown exception caught" << '\n';
+        }
+
+        print_wrong_exception_line(caught, wrong_exception);
```
rianquinn: Have you tested if this works? Seems to me this should print the func / line of the `expect_exception` and not the actual function that is testing the code

connojd: Do you want me to replace EXPECT_* with expect_* in this PR?

rianquinn: no. Lets prove that it works with just the VMCS code to start, and then right after provide a PR that just changes EXPECT, and I would include the expect_true, expect_false with that PR as well.

connojd: Ok will do

connojd:
Updated PR #defines verify_test_results. Each vmcs_ut::test* "calls" verify_test_results to see if they meed the expected result.  The current version prints out the LINE number of vmcs_ut::test* in which _one_ of the tests fails.  There are several actual tests for a single vmcs_ut::test* function (one for each path).  If we want the debug output to indicate which paths failed, the loop in verify_test_results will need to be changed to include a path index.

.travis.yml
```diff
@@ -47,8 +47,8 @@ matrix:
...
-
+<<<<<<< 586aefbff62d898a60f96ad4fa3eb81fda83961a
```
rianquinn: Something is not right here

rianquinn:
Sorry I have been away for so long, how are we doing on this PR?

connojd:
I'm not sure how we can turn EXPECT_TRUE and friends into functions while keeping the \__LINE__ functionality.  For EXPECT_EXCEPTION, I factored out the if-else block that references \__LINE__ into a macro that is used in each unit test (actually it's used in each member of vmcs_ut).

rianquinn:
Yeah, I figured that this would be complicated, so like I said earlier, in this case, I would use a macro with snake case (which is against the Core Guidelines but in this case, it's only being done this way because of __LINE__... if they really wanted that fixed, they would fix the issue like Swift does). 

So this is what we want:

```
this->expect_true(condition);
```

This is how we can implement it:

```
#define expect_true(c) expect_true_with_args(c,__LINE__,__FUNC__)
void expect_true_with_args(bool, int, gsl::string_view)

// Giving us
this->expect_true(condition);
```

Thoughts?

connojd:
Oh yes..that makes sense now that I see it.  The macro actually ends up resolving to a member function.  So would you like for expect_exception to work like that as well?

rianquinn:
might as well keep things consistent.

connojd:
Updated PR has expect_* functions/macros defined in include/unittest.h.  I switched EXPECT_TRUE and EXPECT_FALSE in `test_memory_manager.cpp` to use the `expect_true` and `expect_false` just so that we can see them working without changing them everywhere.

`run_vmcs_test` defined in `bfvmm/src/vmcs/test/test.h` now uses the same technique as the `expect` macros. Examples of `run_vmcs_test` in use are in 
`bfvmm/src/vmcs/test/test_vmcs_intel_x64_check_controls.cpp`

rianquinn:
@connojd Please go through and close the issue tickets that this PR addressed

PR #210, opened 2016-09-15
Updated Memory Manager

rianquinn:
Round three on this. While working on the VMCall interface,
realized that we need a method for managing virtual memory
for memory mapping. The algorithm would be the same for the
memory manager. This patch moves the logic from the memory
manager to a generic memory pool class that can be used
for both the memory manager, but also for managing virtual
memory that will be used for mapping, as well as guest memory
in the future.

In addition, the algorithm was updated to a simple next fit
algorithm. Looking at how heap memory is allocted in our
current baseline, a block size of a CACHE_LINE_SIZE will
cover the vast majority of allocations without fragmentation
issues. Further optimizations would be better accomplished by:
- Creating more than one heap pool with a different block
  size
- Creating custom new / delete operators for classes

### DIFF
--- a/Makefile.bf
+++ b/Makefile.bf
@@ -174,5 +174,5 @@ doxygen_clean:
 
 tidy:
 	@cd %HYPER_ABS%; \
-	%HYPER_ABS%/tools/scripts/verify_source.sh
+	%BUILD_ABS%/build_scripts/verify_source.sh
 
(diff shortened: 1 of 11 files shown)


rianquinn:
@brendank310, @ktemkin If you get a chance, give this PR a look-see.

PR #212, opened 2016-09-19
Don't set -DFORTIFY_SOURCE unless optimizations are on.

ktemkin:
GCC/glibc only correctly fortifies source when optimizations are on; and
defining FORTIFY_SOURCE causes (fatal) warnings on some GCC/glibc versions.
As -DFORTIFY_SOURCE does nothing unless optimizations are on, we'll
only define it for production builds, where optimizations are enabled.

See e.g. https://lists.gnu.org/archive/html/m4-patches/2012-06/msg00002.html for a project with a similar issue.


rianquinn:
LGTM. One note to add is that with the new Clang / LLVM work which is almost done (need to fix TLS first), the above code has been moved to the configure script so that both Native and Cross compilations share flags to keep things consistent. I will remove the fortify call from my PR so that it's consistent with your above fix. 

Thanks for the PR!!!

Issue #193, opened 2016-09-08
operator new emulation and coverity


rianquinn:
The emulation of the new operator to throw bad alloc and provide aligned memory for the VMCS is causing issues with Coverity. To be honest, Clang-Tidy and some of the `-W` warnings were also having issues with the code, so I think it makes sense to see if there is a different approach that these checks aren't as finicky about.

rianquinn:
The PR seems to have addressed this issue. We will need to at some point apply this code to the other unit tests, but so far Coverity doesn't seem to be upset about those, so I am closing this.

PR #221, opened 2016-09-22
Cleanup Coverity Scan Bug

rianquinn:
Coverity scan is getting mad about our use of new / delete
in our unit tests, even though we mimic the standard library
(it's a false positive), so this patch moves this code to a
system header to prevent the scanner from see it. It also
will provide a generic way to do this for other unit tests
in the future.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/193


brendank310:
Skipped the appveyor test as MS servers are rather slow today.

Issue #213, opened 2016-09-19
Windows Broken


rianquinn:
Windows appears to be broken in a couple of different ways:
- Cygwin just did an update which causes more of the GSL code to complain during compilation. We actually contributed strnlen functionality to the GSL specifically for cygwin and it appears that we will need to do the same thing for wcsnlen
- Somehow the IOCTL for write got reversed in our Clang Tidy PR, and as a result, BFM it unable to provide the driver with module information, resulting in BFM failing to load the hypervisor. This is a simple fix that hopefully will not happen again
- The TLS work appears to be killing Windows (BSOD). This doesn't seem to happen as much on real hardware, which allowed this to sneak though during testing as it's a race condition. Specifically, the us of FS seems to make Windows very mad if an interrupt fires while the hypervisor is being loaded. We are investigating different approaches to solve this which include: don't use FS, which would break support for `thread_storage` in the future, or disable interrupts while loading the hypervisor which might be the better approach (if it works). 

In either case, we hope to have Windows support fixed in the very near future.

rianquinn:
I tried disabling interrupts and that didn't help, which is similar to a Linux bug I saw a long time ago, so for now we will have to continue to support interrupts while we launch the VMM. 

So... that means that we cannot touch GS or FS while we load, which is pretty annoying. The Exit Handler can use GS/FS without issue, just not the loader. I will try to draft up a proposal on how I would like to fix the issue today as I believe the TLS implementation is a little buggy how it's implemented today anyways, so I think a new design is in order. 

This all being said, Windows is broken until we get this fixed so I will try to clean this up as fast as possible.

ktemkin:
Looks like it's not just Windows. Picking up Bareflank again, I frequently get a kernel panic while executing pieces of the loader:

```
#make load
<snip>
[  328.419691] [bareflank]: IOCTL_ADD_MODULE_LENGTH: succeeded
[  328.419705] [bareflank]: common_add_module [4]: #intrinsics.so
[  328.419706] [bareflank]:     addr = ffffc900036b9000
[  328.419707] [bareflank]:     size = 0000000000006008
[  328.419707] [bareflank]: IOCTL_ADD_MODULE: succeeded
<snip>
#make start
[  328.858407] BUG: unable to handle kernel paging request at 00007fb969151600
[  328.858510] IP: [<ffffc900036bcea4>] 0xffffc900036bcea4
[...]

PR #222, opened 2016-09-24
Initial Clang Support / Windows Cleanup

rianquinn:
This PR provides some initial support for Clang. There will be a PR in the future to make a change on how clang is called, which will cause another configure version tick, but will keep clang more inline with what we are doing with GCC. 

While developing this patch, I noticed that Windows was completely broken due to two issues documented here:
https://github.com/Bareflank/hypervisor/issues/213

Basically:
- Cygwin did an update that caused more GSL code to stop compiling.
- Windows did an update, which caused some of the driver code to stop compiling, which just needed a swap of some header files
- Our TLS implementation was wrong in so many ways, so I reverted those changes. This could cause stability issues with `on_success` `on_failure`, but for now it seems to work. This code will be fixed shortly in a new PR

In general, this PR should allow us to continue development and testing, including with Clang, while we sort out some issues.


brendank310:
It looks like the clang support puts us beyond the time limit allowed by appveyor for free accounts.

rianquinn:
Yeah, I am working this right now. In the end we will need to download the cross compilers, so this PR will get updated with how to do that. AppVeyor doesn't support docker, so we will basically mimic it instead. 

The harder part is were to put the cross compilers. For now I think I can make it work with DropBox, but we really need to host the file somewhere.

rianquinn:
It's a "festivus miracle". This PR took a long time to get right. Long story short is:

- AppVeyor only gives you a hour, and Clang + GCC compiling on windows takes longer than that. 
- Needed a place to store pre-compiled Windows cross compiler. We now have a file server for this. 
- The cross compilers were originally compiled in Windows 10, which didn't work on Windows 8.1 and 2012 R2, so I had to re-compile in Windows 8.1
- Compilation of BFM on Windows is still GCC, which is fine since we really want to compile BFM and the drivers with Visual Studio at some point anyways.

brendank310:
LGTM

PR #238, opened 2016-10-01
VMCS Field to String

rianquinn:
The following patch cleans up some of the error code logic, and
adds VMCS field to string function so that you get a human
readable output when an error occurs.


brendank310:
Going to close this, as another approach to solving this problem in a cleaner way is being worked on.

PR #243, opened 2016-10-03
Use new expect functions for unit tests

connojd:
This commit replaces each occurence of EXPECT_TRUE,
EXPECT_FALSE, EXPECT_EXCEPTION, and EXPECT_NO_EXCEPTION
in the Bareflank tree with their snake-case counterparts
introduced in (#209).


bfvmm/src/exit_handler/test/test_exit_handler_intel_x64.cpp
```diff
@@ -2446,6 +2449,8 @@ exit_handler_intel_x64_ut::test_vmwrite_failure()
 
     RUN_UNITTEST_WITH_MOCKS(mocks, [&]
     {
-        EXPECT_EXCEPTION(eh->dispatch(), std::runtime_error);
+        auto f = [&] { eh->dispatch(); };
+        auto e = std::shared_ptr<std::exception>(new std::runtime_error("exit handler vmwrite failed"));
```
rianquinn: Why not use `std::make_shared`?

I'd love to see this simplified to something like
```
this->expect_exception([&]{ eh->dispatch() }, std::make_shared<std::runtime_error>("error));
```

What would be even cooler is if there was an overload that allowed us to pass in the function you wanted to call without having to wrap it in a lambda. 

Thoughts?

connojd: I can change to make_shared. Could you give an example of what you would like to see?

connojd: I haven't looked at them all yet, but for the one highlighted, I can use std::bind...it looks like this:

```cpp
this->expect_exception(std::bind(&exit_handler_intel_x64::dispatch, std::move(eh)), std::make_shared<std::runtime_error>("error msg"));
```

rianquinn: Yeah, I'd rather have the lambda 😄 . But the make_shared is much better.

connojd:
I replaced `std::shared_ptr<std::exception>(new ...)` with `std::make_shared`, and std::bind with its lambda equivalent

rianquinn:
Going to close this as it will need some serious work. Once it's cleaned up and ready, @connojd go ahead and resubmit it

Issue #239, opened 2016-10-02
Missing install_clang_1604 function


connojd:
Need to add something like
```bash
install_clang_1604() {
    wget http://llvm.org/releases/3.8.1/clang+llvm-3.8.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz
    tar xf clang*
    sudo cp -R clang*/* /usr/local/
    rm -Rf clang*
}
```
in `tools/scripts/setup_ubuntu.sh` for the 16.04 folks.

rianquinn:
what version ships with Ubuntu 16.04?

rianquinn:
On my system at least, I have 3.8.0-2ubuntu4 which should work fine without having to go through the mess of installing from their site. The only reason I have it in there for 14.04 is clang is too old so I have to update it for Travis. I would note that 14.04 is not actually supported outside of whatever testing Travis is doing for us.

connojd:
The version above works fine and is the one I'm using, but I get a "command not found" error [here](https://github.com/Bareflank/hypervisor/blob/master/tools/scripts/setup_ubuntu.sh#L171) when I try to run setup

rianquinn:
Oh... ha, that just needs to be removed. Feel free to toss in a PR for that if you want.

PR #252, opened 2016-10-09
Fix setup_ubuntu.sh for 16.04

jeanphilippeD:
setup_ubuntu.sh fails because install_clang_1604 is not found.
commit 72ac0c2f94565cfb260fa932d7cf66331d3e4c9 added 2 new functions
to install clang 'install_clang' (use apt), 'install_clang_1404' (use tar),
but call install_clang_1604.

Fix:
-Add install_clang_1604 similar to install_clang_1404.
-does not use install_clang as apt seems to have a slightly different package
 3.8 instead of 3.8.1 from the tar in 14.04

Test:
make -j2 complete successfully

### DIFF
--- a/tools/scripts/setup_ubuntu.sh
+++ b/tools/scripts/setup_ubuntu.sh
@@ -111,6 +111,13 @@ install_clang_1404() {
     rm -Rf clang*
 }
 
+install_clang_1604() {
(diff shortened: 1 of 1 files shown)


rianquinn:
Thanks you. This fixed the following: https://github.com/Bareflank/hypervisor/issues/239

PR #263, opened 2016-10-18
bfm: Optimized execution time for split function

RicoAntonioFelix:
Signed-off-by: Rico <“ricoantoniofelix@yahoo.com”>


rianquinn:
Merging as this is a much better implementation and I would like this in our git history in case we need to do something similar in the future. Since split is no longer needed in BFM it will be removed in future PRs. 

Thanks Rico!!!

PR #269, opened 2016-10-22
Add 32-bit vmcs control namespace #1

connojd:
This commit moves every 32-bit vmcs control up to the
VM-entry instruction length to their own namespaces.

I've also added some functions to the ctls msrs to aid in the control checks.  In a later PR, I plan to cleanup the vmcs test files whose lines are a little verbose.  For example, `enable_proc_ctl`, `disable_proc_ctl`, etc aren't really necessary now so they will eventually be removed in favor of the namespace code.


bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -29,6 +29,13 @@
 #include <intrinsics/vmx_intel_x64.h>
 #include <intrinsics/msrs_intel_x64.h>
 
+extern void
```
rianquinn: Not a huge fan of these being here. Also `extern` should be dropped

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -1917,20 +1829,1860 @@ constexpr const auto VMCS_HOST_IA32_PERF_GLOBAL_CTRL                      = 0x00
...
+    namespace external_interrupt_exiting
+    {
+        constexpr const auto mask = 0x00000001UL;
+        constexpr const auto from = 0;
+        constexpr const auto name = "external_interrupt_exiting";
+
+        inline bool
+        is_enabled()
+        { return (vm::read(addr, name) & mask) != 0; }
+
```
rianquinn: Each control should have it's own `set` and `set_if_allowed` functions and the `enable`/`disable` and friends should use the `set` functions. There should be no need to pass in the msrs as you already have the `mask` (they are the same... that is... the VMCS controls and the MSRs use the same bit masks, so there is no need to use the verbose version). 

In general, I would really like to simplify the  set_vm_control function. To do that, you might want to add a `constexpr const auto addr_msr = msrs::<control>::addr. This way, you have the addr, you have the mask, you should be able to simplify this function so that it's a cut / paste on all of the controls.

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -33,8 +33,55 @@
 using namespace x64;
 using namespace intel_x64;
 
-vmcs_intel_x64::vmcs_intel_x64() :
-    m_vmcs_region_phys(0)
+void
+set_vm_control(uint32_t val, bool is_allowed0, bool is_allowed1,
```
rianquinn: I would write this using template coding so that it's not so type specific, and so that it's definition can be in the .h. Also, it will likely change slightly with the comments above

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -33,8 +33,55 @@
...
+               uint64_t addr, const std::string &name, uint64_t mask)
+{
+    using namespace intel_x64::vmcs;
+
+    if (val == 0)
+    {
+        if (is_allowed0)
```
rianquinn: Swap the order on this and drop the else. I'm actually surprised Clang Tidy didn't complain about that. But in general, if / else should be avoided as per the readability checks.

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -222,19 +269,19 @@ vmcs_intel_x64::write_32bit_control_state(const std::shared_ptr<vmcs_intel_x64_s
 
     lower = ((ia32_vmx_pinbased_ctls_msr >> 0) & 0x00000000FFFFFFFF);
     upper = ((ia32_vmx_pinbased_ctls_msr >> 32) & 0x00000000FFFFFFFF);
-    vm::write(VMCS_PIN_BASED_VM_EXECUTION_CONTROLS, lower & upper);
+    vmcs::pin_based_vm_execution_controls::set(lower & upper);
```
rianquinn: So pretty 😄

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -496,39 +543,39 @@ vmcs_intel_x64::secondary_processor_based_vm_execution_controls()
...
-    controls |= VM_EXIT_CONTROL_LOAD_IA32_PERF_GLOBAL_CTRL;
-    controls |= VM_EXIT_CONTROL_ACKNOWLEDGE_INTERRUPT_ON_EXIT;
-    controls |= VM_EXIT_CONTROL_SAVE_IA32_PAT;
-    controls |= VM_EXIT_CONTROL_LOAD_IA32_PAT;
-    controls |= VM_EXIT_CONTROL_SAVE_IA32_EFER;
-    controls |= VM_EXIT_CONTROL_LOAD_IA32_EFER;
-    // controls |= VM_EXIT_CONTROL_SAVE_VMX_PREEMPTION_TIMER_VALUE;
+    auto controls = vmcs::vm_exit_controls::get();
+
+    controls |= vmcs::vm_exit_controls::save_debug_controls::mask;
```
rianquinn: These all need to be converted. The approach here is just wrong (both the old one and this one). Instead, each control should be set using the `enable` function with the exception to this being the could in the secondary controls like XSAVE, RDTSCP and the invalidation instruction which should be `enable_if_allowed`.

Just like I had before, enable all of them using the same scheme, and then comment out the ones we don't use.

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -496,39 +543,39 @@ vmcs_intel_x64::secondary_processor_based_vm_execution_controls()
...
-    controls |= VM_ENTRY_CONTROL_LOAD_IA32_EFER;
+    controls |= vmcs::vm_entry_controls::load_debug_controls::mask;
+    controls |= vmcs::vm_entry_controls::ia_32e_mode_guest::mask;
+    // controls |= vmcs::vm_entry_controls::entry_to_smm::mask;
+    // controls |= vmcs::vm_entry_controls::deactivate_dual_monitor_treatment::mask;
+    controls |= vmcs::vm_entry_controls::load_ia32_perf_global_ctrl::mask;
+    controls |= vmcs::vm_entry_controls::load_ia32_pat::mask;
+    controls |= vmcs::vm_entry_controls::load_ia32_efer::mask;
 
     this->filter_unsupported(msrs::ia32_vmx_true_entry_ctls::addr, controls);
```
rianquinn: Ditch this function.

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
```diff
@@ -637,63 +615,57 @@ vmcs_intel_x64::check_control_event_injection_delivery_ec_checks()
...
         return;
 
-    if ((exception_error_code & 0x00000000FFFF8000) != 0)
+    if ((vmcs::vm_entry_exception_error_code::get() & 0x00000000FFFF8000UL) != 0)
         throw std::logic_error("bits 31:15 of the exception error code field must be 0 "
                                "if deliver error code bit is set in the interrupt info field");
 }
 
 void
 vmcs_intel_x64::check_control_event_injection_instr_length_checks()
```
rianquinn: The way you coded this is awesome. I never thought to use the `using` feature to qualify a type, which is pretty badass. Great job here!!!

bfvmm/src/vmcs/src/vmcs_intel_x64_check_guest.cpp
```diff
@@ -1492,7 +1492,7 @@ vmcs_intel_x64::check_guest_rflags_reserved_bits()
 void
 vmcs_intel_x64::check_guest_rflags_vm_bit()
 {
-    if (!is_enabled_ia_32e_mode_guest() && vmcs::guest_cr0::protection_enable::get() == 1)
+    if (!vmcs::vm_entry_controls::ia_32e_mode_guest::is_enabled() && vmcs::guest_cr0::protection_enable::get() == 1)
```
rianquinn: Look for instances of this where the one line if statement is compound and just turn them into two separate checks for two reasons:
- This ensures coveralls is checking everything (one liners can hide one of the checks)
- it's too verbose

bfvmm/include/vmcs/vmcs_intel_x64.h
rianquinn: These should be right above the VMCS fields as a helper, written using generic programming defined int he header and not in the .cpp

bfvmm/include/vmcs/vmcs_intel_x64.h
connojd: Thats what I was doing originally.  I thought that having the definition of the set_* functions in one place would help in the event of changes to it...instead of having to change the definition in every control you only have to do it in one place.  I guess we'll leave it up to code review and testing to make sure all the definitions match

rianquinn: I think if you have a addr_msr, the code will be cut / paste (literally it will be the same for every control) which too me is better so that if you have to make a change, it can be done quickly for all of them.

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
connojd: I'm going to ypdate all of these in the next PR with the changes for the secondary controls.

Issue #267, opened 2016-10-22
VMCS launch failure debug output not working


rianquinn:
If a VMCS launch failure occurs, we get the following:

```
ERROR: vmlaunch failed:
ERROR:     - vm_instruction_error: VM entry with invalid host-state field(s)
ERROR: vm::read failed:
ERROR:     - field:
std::exception caught in ~final_act: vm::read failed

ERROR: ----------------------------------------
ERROR: - Standard Exception Caught            -
ERROR: ----------------------------------------
ERROR: vm::launch failed
```

The vmread should not be failing, and even if it did, it should tell us which field failed (likely not being reported due to incomplete transition to new VMCS code, see [here](https://github.com/Bareflank/hypervisor/issues/241)). Once this transition is done, we need to test VMCS launch failures to ensure the output makes sense, is easy to read, and works as expected.

rianquinn:
@connojd Please look into this

rianquinn:
Was was due to the PML address being read before support was identified. Was fixed in #271

PR #271, opened 2016-10-28
Fix borken commit

rianquinn:
The unit tests are passing but the hypervisor was broken in
this last commit. This patch fixes the problem


rianquinn:
Fixed: #267

PR #272, opened 2016-10-28
VM controls namespace implementation

connojd:
This completes the move of the execution/entry/exit controls
to namespaces.


bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -3453,187 +3380,794 @@ namespace guest_es_access_rights
...
-    {
-        constexpr const auto mask = 0x0000000000001000UL;
-        constexpr const auto from = 12;
-        constexpr const auto name = "avl";
+        template <class T> void set(T val)
+        { set_vm_control(val, msr_addr, addr, name, mask); }
 
-        inline auto get()
-        { return (vm::read(addr, name) & mask) >> from; }
+        template <class T> void set_if_allowed(T val, bool verbose) noexcept
```
rianquinn: The diff is really hard to read, but did we still have a get? I cannot remember if we talked about that. We have a set, and set_if_allowed, we probably should have the same thing for get and get_if_allowed

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -3453,187 +3380,794 @@ namespace guest_es_access_rights
...
+    }
+
+    namespace enable_vpid
+    {
+        constexpr const auto mask = 0x0000000000000020UL;
+        constexpr const auto from = 5;
+        constexpr const auto name = "enable_vpid";
+
+        inline bool
+        is_enabled()
```
rianquinn: Yeah, now that I look at this, we really need a get and get_if_enabled. And then is_enabled and is_enabled_if_allowed should use get and get_if_allowed. Otherwise, the is_enabled as it written would result in an exception. The dump() function in the VMCS currently uses a macro which will not be needed once this work is done, but if_allowed is really important, otherwise attempting to dump while an exception was thrown will result in yet another exception.

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
```diff
@@ -77,7 +79,16 @@ vmcs_intel_x64::check_control_ctls_reserved_properly_set(uint64_t msr_addr, uint
         throw std::logic_error(std::string("invalid ") + name);
     }
 
-    if ((ctls & ~allowed1) != 0UL)
+    allowed1_failed = (ctls & ~allowed1) != 0UL;
+
+    if (msrs::ia32_vmx_procbased_ctls2::addr == msr_addr)
+    {
+        using namespace vmcs::primary_processor_based_vm_execution_controls;
+
```
rianquinn: Extra while space is not needed, here. Also `bool allowed1_failed = false` is redundant. Prefer `auto allowed1_failed = false`, and I think the line below can be simplified with `&=`, and if we are doing a ton of `using namespace` stuff, we should just qualify those namespaces at the top of the file and call it a day 😄

bfvmm/include/vmcs/vmcs_intel_x64.h
connojd: Right now there is a `get()`, `set()`, and `exists()` in the field's namespace.  For fields that are controls, each control's namespace has `is_enabled()`, `enable`, `disable`, `enable_if_allowed`, `disable_if_allowed`. The last four call `set()` and `set_if_allowed`. Those are also in the control's namespace.

bfvmm/include/vmcs/vmcs_intel_x64.h
connojd: Let me make sure I know what you want. For the control field's namespace like `secondary_processor_vm_execution_controls`, `get_if_allowed` needs to be added.  Then  each control's namespace within the field (like enable_vpid) will have an `is_enabled` which calls `secondary_processor_based_vm_execution_controls::get` and an `is_enabled_if_allowed` which calls `secondary_processor_based_vm_execution_controls::get_if_allowed`?

rianquinn: Correct

connojd: Do you think the `get` should throw an exception?  Something like
```cpp
namespace pin_based_vm_execution_controls
{
    inline auto get()
    { 
        if (!exists())
            throw std::logic_error("pin based controls don't exist");
     
        return vm::read(addr, name);
    }

    inline auto get_if_allowed(bool verbose = false)
    {
        if (!exists() && verbose)
            bfwarning << name << " doesn't exist" << '\n';
 
        if (exists())
            return vm::read(addr, name);
    }
}
```
Another question...do you think we need these for every field, or just the ones that may not exist? I'm pretty sure the pin based controls and primary controls are going to exist on every processor we support, so in those cases I think a  `get_if_allowed` would be redundant

Also we should probably make a `set_if_allowed` to go along with `get_if_allowed` as well.

rianquinn: We should do them for all of them so that there is consistency. For some of them yes, `exists()` will just return true, but I don't want people to have to figure out what is what (more room for error). And yes, your implementation is perfect, and yes we need a set_if_allowed (I actually thought we did as that was being used for enable_if_allowed and friends).

connojd:
@rianquinn check `namespace pin_based_vm_execution_controls` and its sub-namespaces and let me know what you think. I haven't included the tests yet because I want to make sure I'm headed in the right direction.

A couple of thoughts/questions:
This may be a dead horse by know but are `set` and `set_if_allowed` in the namespaces which also have `enable`, `enable_if_allowed`, etc. necessary?  With the way they are written right now, we could just call `set_vm_control` and `set_vm_control_if_allowed` directly from `enable` and friends.  This would make the intent of the control interface more clear and would also make it harder to abuse.  If the user really wanted to use a value other than 0 or 1, they still could by using the field's set function even though that wouldn't make sense.

I haven't looked for others yet, but the vmx_procbased_ctls2 exists iff the secondary proc ctls field exists in the vmcs..should we be checking for msrs' existence as well? At least in the vmcs code, the existence of a field (like secondary proc ctls) implies the existence of its underlying msr, so I think we could get by with not checking for the msrs since we'll already be checking for the fields.

rianquinn:
LGTM. 

I agree that maybe set is probably not needed, but then neither would `get` be too (you could always create a `get_vm_control` as well. Consistency is all I really care about here. So at a minimum, look into whether you we could use a `get_vm_control`. 

But yeah, I guess the following would work, and provide more than enough, and it would keep things simple. Since the `disabled` versions are just the reverse, they likely are not needed either. 

```cpp
is_enabled() // run get_vm_control
is_enabled_if_allowed() // run get_vm_control_if_allowed
enable() // run set_vm_control
enable_if_allowed() // run set_vm_control_if_allowed
```

Thoughts?

connojd:
I think the disable and disable_if_allowed is still needed for clearing the control to 0

rianquinn:
good point, in which case I would just leave in the is_disabled and friends for consistency

rianquinn:
so

```cpp
is_enabled() // run get_vm_control
is_enabled_if_allowed() // run get_vm_control_if_allowed
is_disabled() // run get_vm_control
is_disabled_if_allowed() // run get_vm_control_if_allowed
enable() // run set_vm_control
enable_if_allowed() // run set_vm_control_if_allowed
disable() // run set_vm_control
disabled_if_allowed() // run set_vm_control_if_allowed
```

Thoughts?

connojd:
Looks good :+1:

connojd:
I changed the function names slightly to reflect the manual better. Each vmcs field has get, set, get_if_exists, and set_if_exists.  The control fields' sub namespaces have is_enabled, is_disabled, is_enabled_if_exists, is_disabled_if_exists, enable, disable, enable_if_allowed, and disable_if_allowed.

rianquinn:
Love it, let me know when you update the PR and I will re-do the review.

connojd:
Just pushed it

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -454,37 +454,28 @@ vmcs_intel_x64::primary_processor_based_vm_execution_controls()
...
-
-    vm::write(VMCS_SECONDARY_PROCESSOR_BASED_VM_EXECUTION_CONTROLS, controls);
+    using namespace vmcs::secondary_processor_based_vm_execution_controls;
+
+    bool verbose = true;
+
+    // virtualize_apic_accesses::enable_if_allowed(verbose);
+    // enable_ept::enable_if_allowed(verbose);
+    // descriptor_table_exiting::enable_if_allowed(verbose);
+    enable_rdtscp::enable_if_allowed(verbose);
```
rianquinn: Shouldn't this be changed to `exists`?

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
```diff
@@ -454,37 +454,28 @@ vmcs_intel_x64::primary_processor_based_vm_execution_controls()
...
-    // and SXAVE should all be uncommented, with the rest commented.
-    // I commented out xsave for now so that the hypervisor will still
-    // work on VMWare for now.
-    //
-    // controls |= VM_EXEC_S_PROC_BASED_ENABLE_XSAVES_XRSTORS;
-
-    this->filter_unsupported(msrs::ia32_vmx_procbased_ctls2::addr, controls);
-
-    vm::write(VMCS_SECONDARY_PROCESSOR_BASED_VM_EXECUTION_CONTROLS, controls);
+    using namespace vmcs::secondary_processor_based_vm_execution_controls;
```
rianquinn: Also, for all of these in the VMCS, if you could, get rid of the using namespace in these functions. Qualify the `vmcs` namespace at the top with `x64` and `intel_x64` and leave the secondary_processor_based_vm_execution_controls and friends on each call. I know it will make the code a little longer, but I have been seeing it that way for almost a year, and it's easier for to spot each control with the duplication 😄

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
```diff
@@ -361,12 +384,19 @@ vmcs_intel_x64::check_control_enable_ept_checks()
...
 {
-    if (!is_enabled_pml())
+    using namespace vmcs::primary_processor_based_vm_execution_controls;
+    using namespace vmcs::secondary_processor_based_vm_execution_controls;
+
+    bool secondary_ctls = activate_secondary_controls::is_enabled();
+    bool pml = enable_pml::is_enabled();
+    bool ept = enable_ept::is_enabled();
+
+    if (!pml || !secondary_ctls)
```
rianquinn: There are a couple of things to do here:

- Both PML and EPT should be `is_enabled_if_exists()`. Without that, attempting to read the control will fail. We can assume the secondary controls exists since we require the TRUE based controls which also support the secondary. 
- The `||` should be broken up into two tests. 
- Drop the extra variables as they should not be needed. Also... check to make sure you are not defining `bool` anywhere else as it should be `auto`... otherwise your defining the type twice. 
- I would drop the using namespace here and use the full name as you should have more room once you move to two separate tests

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
```diff
@@ -379,17 +409,27 @@ vmcs_intel_x64::check_control_enable_pml_checks()
...
 void
 vmcs_intel_x64::check_control_enable_vm_functions()
 {
-    if (!is_enabled_vm_functions())
+    using namespace vmcs::primary_processor_based_vm_execution_controls;
+    using namespace vmcs::secondary_processor_based_vm_execution_controls;
+
+    bool secondary_ctls = activate_secondary_controls::is_enabled();
+
+    if (!enable_vm_functions::is_enabled() || !secondary_ctls)
```
rianquinn: same thing here... two different tests, no `bool` and should be using `if_exists()` versions for all of the secondary ones. 

The reality is... your going to have to do through these checks and make sure that you clean the above up as I bet there are several spots where this is messed up. And likely... you can just use the `if_exists` for most of these to keep it clean.

include/unittest.h
```diff
@@ -673,8 +673,6 @@ class unittest
             this->expect_failed(e.what(), func, line);
         }
 
-        //inc_pass();
```
rianquinn: Why is this commented out?

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
connojd: you got it :+1:

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
connojd: allowed implies exists the way its written right now

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
connojd: Hmm I thought I already did that... ok ill do a sweep and clean it up

bfvmm/src/vmcs/src/vmcs_intel_x64.cpp
rianquinn: Based on what you stated above, I just assumed that you were getting rid of the `allowed` in favor of `exists` so I guess I'm confused / surprised this even compiles

connojd: No, each control like enable_rdtscp has enable, disable, enable_if_allowed, disable_if_allowed, is_enabled_if_exists, is_disabled_if_exists, is_enabled, is_disabled

rianquinn: Ok.... so this is where the issue is. 

There is a difference between allowed, and exists. The MSRs state if a control is allowed to be 0 or 1. You can actually have a control that exists, but is not allowed to be 0 or 1. A good example of that is exit on write to CR3. That has always existed, but was not allowed to be 0 until the TRUE based controls came out. 

So I like your naming scheme, as it handles both cases well. You can only enable / disable if it is both allowed and exists, and you can only query is it exists (allowed doesn't matter).

So +1

RicoAntonioFelix: Logic will state that <something> can exist in reality but not allowed for {some entities} to know of its existence, in order for {some entities} to know about <something>'s existence it must first exist and then be allowed...

bfvmm/src/vmcs/src/vmcs_intel_x64_check_controls.cpp
connojd: or everywhere?

rianquinn: I would drop it where it makes sense. If you end up with a line that is huge, obviously keep it, but I like having the full name there as I don't have to figure out which namespace it came from.

connojd:
@rianquinn  i've got the checks cleaned up.

connojd:
Ah crap I need to change the control functions in vmcs_intel_x64..one moment

connojd:
@rianquinn ok ready for review

PR #276, opened 2016-11-03
Update 16-bit guest field namespace

connojd:
Updates the 16-bit guest state fields to use if_exists.

### DIFF
--- a/bfvmm/include/vmcs/vmcs_intel_x64.h
+++ b/bfvmm/include/vmcs/vmcs_intel_x64.h
@@ -686,14 +686,20 @@ namespace guest_es_selector
     constexpr const auto addr = 0x0000000000000800UL;
     constexpr const auto name = "guest_es_selector";
 
-    inline auto get()
-    { return vm::read(addr, name); }
-
-    template<class T> void set(T val)
-    { vm::write(addr, val, name); }
-
     inline bool exists() noexcept
     { return true; }
+
+    inline auto get()
+    { return get_vmcs_field(addr, name, exists()); }
+
+    inline auto get_if_exists(bool verbose = false) noexcept
+    { return get_vmcs_field_if_exists(addr, name, verbose, exists()); }
+
+    template<class T> void set(T val)
+    { set_vmcs_field(val, addr, name, exists()); }
+
+    template<class T> void set_if_exists(T val, bool verbose = false) noexcept
+    { set_vmcs_field_if_exists(val, addr, name, verbose, exists()); }
 
     namespace rpl
     {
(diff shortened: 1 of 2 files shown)


bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -728,10 +758,22 @@ namespace guest_es_selector
...
-        { return (vm::read(addr, name) & mask) >> from; }
+        { return (get_vmcs_field(addr, name, exists()) & mask) >> from; }
+
+        inline auto get_if_exists(bool verbose = false) noexcept
+        { return (get_vmcs_field_if_exists(addr, name, verbose, exists()) & mask) >> from; }
 
         template<class T> void set(T val)
-        { vm::write(addr, (vm::read(addr, name) & ~mask) | ((val << from) & mask), name); }
+        {
+            auto field = get_vmcs_field(addr, name, exists());
```
rianquinn: This is fine because it's an int (best optimization), but for future reference, please watch this, and if you find any issues with what we have done, please correct them. 

https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -568,6 +568,10 @@ vmcs_ut::test_vmcs_guest_es_selector()
 
     this->expect_true(vmcs::guest_es_selector::get() == 100UL);
     this->expect_true(vmcs::guest_es_selector::exists());
+
+    vmcs::guest_es_selector::set_if_exists(200UL);
+
```
rianquinn: remove whitespace

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -607,6 +629,10 @@ vmcs_ut::test_vmcs_guest_cs_selector()
 
     this->expect_true(vmcs::guest_cs_selector::get() == 100UL);
     this->expect_true(vmcs::guest_cs_selector::exists());
+
+    vmcs::guest_cs_selector::set_if_exists(200UL);
+
```
rianquinn: whitespace.. look for these and get rid of the rest

bfvmm/src/vmcs/test/test_vmcs_intel_x64.cpp
```diff
@@ -871,17 +1041,28 @@ vmcs_ut::test_vmcs_guest_tr_selector_index()
...
+    vmcs::guest_tr_selector::index::set_if_exists(0UL);
+    this->expect_true(vmcs::guest_tr_selector::index::get_if_exists() == 0UL);
 }
 
 void
 vmcs_ut::test_vmcs_guest_interrupt_status()
 {
-    vmcs::guest_interrupt_status::set(100UL);
-    g_msrs[msrs::ia32_vmx_procbased_ctls2::addr] = msrs::ia32_vmx_procbased_ctls2::virtual_interrupt_delivery::mask;
+    g_msrs[msrs::ia32_vmx_true_procbased_ctls::addr] = msrs::ia32_vmx_true_procbased_ctls::activate_secondary_controls::mask << 32;
```
rianquinn: if exists relies on the secondary controls, we should be testing when they are disabled as well as enabled, and the same goes for the interrupt status. 

Please review other fields and make sure that these same tests are not missing

bfvmm/include/vmcs/vmcs_intel_x64.h
connojd: Will do; that is a great article btw

connojd:
@rianquinn I've made the changes. Also went back to the 16bit control fields that depend on secondary controls and tested when they're disabled.

Issue #274, opened 2016-11-02
Throwing from the exit handler crashes the system


rianquinn:
Not sure why, but if you throw from within the exit handler, the system crashes with a triple fault. Interestingly enough, the code never makes it the unwinder, and the same code works fine if thrown while bringing up the hypervisor, so it's only an issue with the exit handler. It will crash on anything that is thrown too. 

My current hunch is that we have something in hardware disabled that exception support needs, or there is an issue with TLS or something like that which is going wrong. Best way to debug this from what I can tell is diving into libcxxabi and seeing how far it's getting.

rianquinn:
This ended up being an issue with the TLS code, which also effected our use of std::uncaught_exceptions (so it was more than just throwing). The driver was never providing the TLS region to the VMM's memory manager, so there were no page entries for this memory. As soon as libc++ attempted to access the TLS data from within the exit handler, a triple fault would occur. This will be fixed in the upcoming VMCall PR.

Issue #182, opened 2016-09-06
[RFC]: VMCall Interface


rianquinn:
In the next couple of days we will be implementing a VMCall interface that can be used by users of Bareflank to quickly and easily talk to the hypervisor. Our goal here is to keep this as simple as possible so that it can be extended in the future without causing issues with API changes

Since we intended to work with ARM in the future, we cannot VMCall from userspace as this is not supported on ARM (sad story). Instead, we will provide an IOCTL interface so that userspace applications like BFM can VMCall to the hypervisor. 

One issue with supporting both ARM and Intel is the total number of registers. To support this in a sane way, I propose we use the ABIs are our guide. On Intel, the System V ABI defines the order of each register. We can define a struct that has 32 registers (i.e. r0 -> r31) and the ABI will tell use which register in our struct belongs to which physical register (i.e. r0 -> rax, r1 -> rdx, etc...). From here, the driver can use this information to produce a VMCall, and the only "arch" specific code will be the register decoding. This should produce a simple VMCall interface that can be used generically. 

In addition to register decoding, Bareflank will need to reserve a register for defining the type of VMCall that is occurring. At the moment, I envision three different types of VMCalls, but this could be expanded in the future:
- raw register VMCall (minus reserved registers)
- data VMCall
- unit test VMCall

The raw register VMCall would do nothing more than send raw register values to the hypervisor. This is likely to be the most used VMCall. The data VMCall would send in / out pages to the hypervisor to handle. The unit test VMCall can be used to unit test code inside the hypervisor (e.g. libcxx). To ensure future compatibility, I would like to reserve r0 and r1 (rax / rdx on Intel) for Bareflank, and extensions can use all of the registers for whatever they want. 

Thoughts?

To start, I would like to add raw register and unit test support to BFM, and the data VMCall can be added later (its a little more complicated as it has to map / unmap memory, and we need a clean way to send binary / string data via BFM).

brendank310:
Overall, this sounds pretty reasonable to me; creating a high and low bandwidth option for exchanging data with the hypervisor is going to be required for all non-trivial feature additions. I see preventing calls from userspace to be a feature, to allow whatever policy enforcement over vmcall access that exists in the operating system to be leveraged. I don’t agree that unit tests is a separate category of vmcall, it more than likely falls into one of the other two categories, for different test cases. But if the statement about the 3 types is about reserving the values for the functional use of the register, page sharing, and unit tests I have no objections.

[...]

PR #281, opened 2016-11-08
VMCall Interface

rianquinn:
This provides the support for 'vmcall'. BFM has been modified
to support the vmcall interface, but custom applications can
use this interface to do whatever they want, including calling
vmcall from user-space if so desired.

This patch also provides new code in the memory manager to
handle mapping memory from the Host OS into the hypervisor.

Finally, this patch fixes a host of bugs including a bug
with TLS, and cleans up the code that was touched in
prep for 1.1

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/182
[ISSUE]: https://github.com/Bareflank/hypervisor/issues/274


bfvmm/include/exit_handler/exit_handler_intel_x64.h
```diff
@@ -135,16 +145,23 @@ class exit_handler_intel_x64
     virtual void handle_xsaves();
     virtual void handle_xrstors();
 
-    virtual void advance_rip();
-    virtual void unimplemented_handler();
+    virtual void advance_rip() noexcept;
+    virtual void unimplemented_handler() noexcept;
```
brendank310: I can understand noexcept for advance_rip, but is there not a case where unimplemented would/should throw?

bfvmm/include/memory_manager/map_ptr_x64.h
```diff
@@ -0,0 +1,876 @@
...
+            auto &&vmap = upper(virt);
+            for (auto vadr = vmap; vadr < vmap + size; vadr += page_size)
+                g_pt->unmap(vadr);
+
+            g_mm->free_map(reinterpret_cast<pointer>(vmap));
+        }
+    }
+
+    integer_pointer
+    virt_to_phys_wuth_cr3(integer_pointer virt, integer_pointer phys)
```
brendank310: s/wuth/with

bfvmm/include/memory_manager/page_table_entry_x64.h
```diff
@@ -59,155 +76,257 @@ class page_table_entry_x64
...
     /// Set Dirty
     ///
+    /// @expects none
+    /// @ensures none
+    ///
     /// @param enabled true if the entry is dirty, false otherwise
     ///
     void set_dirty(bool enabled) noexcept;
 
+    /// Page Size
```
brendank310: This seems kind of odd naming convention.

bfvmm/include/vmcs/vmcs_intel_x64_host_vm_state.h
```diff
@@ -109,77 +111,77 @@ class vmcs_intel_x64_host_vm_state : public vmcs_intel_x64_state
...
         bfdebug << "segment selectors:" << bfendl;
-        PRINT_STATE(m_es);
-        PRINT_STATE(m_cs);
-        PRINT_STATE(m_ss);
-        PRINT_STATE(m_ds);
-        PRINT_STATE(m_fs);
-        PRINT_STATE(m_gs);
-        PRINT_STATE(m_ldtr);
-        PRINT_STATE(m_tr);
+        bfdebug << std::setw(35) << view_as_pointer(m_es) << bfendl;
```
brendank310: 35?

bfvmm/src/exit_handler/src/exit_handler_intel_x64.cpp
```diff
@@ -934,3 +1028,132 @@ exit_handler_intel_x64::exit_reason_to_str(uint64_t exit_reason)
             return "unknown";
     };
 }
+
+void
+exit_handler_intel_x64::handle_vmcall_versions(vmcall_registers_t &regs)
+{
+    switch (regs.r02)
+    {
+        case 0:
```
brendank310: Label the cases?

include/constants.h
```diff
@@ -23,6 +23,38 @@
...
+#endif
+#endif
+
+/*
+ * Hypervisor Version
+ *
+ * Uses http://semver.org
+ */
+#define BAREFLANK_VERSION_MAJOR 1ULL
+#define BAREFLANK_VERSION_MINOR 1ULL
```
brendank310: Might want to be careful setting the version before the tag. This type of versioning can cause problems in the future.

include/constants.h
```diff
@@ -87,10 +119,10 @@
  * anything that uses a std::container uses this heap so it does need to
  * have some size to it. Pages do not come from this pool.
  *
- * Note: defined in bytes
+ * Note: defined in bytes (defaults to 8MB)
  */
 #ifndef MAX_HEAP_POOL
-#define MAX_HEAP_POOL (256ULL * MAX_PAGE_SIZE * MAX_CACHE_LINE_SIZE)
+#define MAX_HEAP_POOL (256ULL * MAX_PAGE_SIZE * 8)
```
brendank310: Maybe make a define for the 8, so heap size can be tuned easily.

include/constants.h
```diff
@@ -99,13 +131,37 @@
  * This defines the internal memory that the hypervisor allocates to use
  * for allocating pages.
  *
- * Note: defined in bytes (defaults to 8MB)
+ * Note: defined in bytes (defaults to 1MB)
```
brendank310: Same, let's make this tunable.

include/maskmanip.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <type_traits>
+
+template<class T, class M,
+         class = typename std::enable_if<std::is_integral<T>::value>::type,
+         class = typename std::enable_if<std::is_integral<M>::value>::type>
+auto get_mask(T t, M m) noexcept
+{ return t & m; }
+
+template<class T, class M,
```
brendank310: Both clear and set mask aren't particularly intuitive. They return a mask, not actually set/clear anything on a particular mask.

bfvmm/include/exit_handler/exit_handler_intel_x64.h
rianquinn: no, we want unimplemented to gracefully call halt when it's done

bfvmm/include/memory_manager/page_table_entry_x64.h
rianquinn: Yeah, that's out of the Intel manual. PS == Page Size, and it basically means, does this represent a page (i.e. this page size == #bits) or does this represent another set of page tables (i.e. page size - bits). I hate it too

bfvmm/include/vmcs/vmcs_intel_x64_host_vm_state.h
rianquinn: 35 because everything fits with that width. It's easy to find/replace if we want to change the alignment to something else. It was actually there all along, all I did was remove the macro (as that's a general thing that will be happening in coming PRs before 1.1)

bfvmm/src/exit_handler/src/exit_handler_intel_x64.cpp
rianquinn: https://github.com/rianquinn/hypervisor/blob/vmcall_interface/include/vmcall_interface.h#L81

rianquinn: I could add an enum to this interface if you would like, but they are intended to be arbitrary as the user can add to this as they wish

include/constants.h
rianquinn: Yeah, we are about to roll 1.1 so I think this is fine for now, but in the future, it will stay on the current tag right up to the point where it is tagged to a new version

include/constants.h
rianquinn: no #define... just sizeof(uintptr_t)

include/constants.h
rianquinn: samething... sizeof(uintptr_t), it's already tunable by changing the #define on the command line (hence why each of these have a #ifdef

include/maskmanip.h
rianquinn: What naming would you prefer?

bfm/src/arch/linux/ioctl.cpp
```diff
@@ -22,82 +22,72 @@
 #include <ioctl.h>
 #include <ioctl_private.h>
 
-ioctl::ioctl() noexcept
-{
-    m_d = std::make_shared<ioctl_private>();
-}
+ioctl::ioctl() noexcept :
+    m_d {std::make_unique<ioctl_private>()}
```
RicoAntonioFelix: I'm not to sure if this constructor should be marked `noexcept` as `std::make_unique` has potential to throw `std::bad_alloc` or if `ioctl_private`'s constructor throws will violate the `noexcept` contract...

rianquinn:
Changes are in.

PR #302, opened 2016-11-11
Finish 32-bit vmcs control fields

connojd:
This updates the 32-bit vmcs control fields to use
std::enable_if and bitmanip.h.


connojd:
@rianquinn formatting has been updated in all 16-bit fields and 32-bit control fields

PR #303, opened 2016-11-11
Fix problem with vmcall interface that was accessing user space memor…

brendank310:
…y directly.


rianquinn:
This implementation has two problems. The malloc could leak in the event of an error and its slow. Please convert to stack based.

brendank310:
I will give you the leak, but the speed concern is pretty ridiculous and speculative.

rianquinn:
LGTM

PR #312, opened 2016-11-15
Remove overuse of shared_ptr

rianquinn:
We are using shared_ptr too much when it's really not needed.
This not only adds overhead, but it is misleading. This patch
removes all but a couple instances of shared_ptr where it is
not needed. A future patch will remove the remaining uses
in an attempt to keep this patch smaller.


rianquinn:
Hold off on this PR until i update it. The other two from me look good.

rianquinn:
@brendank310 this has been fixed. It should be good to merge once Travis CI is working again

PR #315, opened 2016-11-15
rm std string

connojd:
This removes the use of std::string in namespace vmcs and
namespace vm in order to prevent freezing of the system.

I believe the source of the freeze problem is too high
a perf sampling rate.  The lower the perf_cpu_time_max_percent
is, the less time is spent sampling NMIs. Disabling it
completely makes the system act as if bareflank isn't running at all.

### DIFF
--- a/bfvmm/include/vmcs/vmcs_intel_x64.h
+++ b/bfvmm/include/vmcs/vmcs_intel_x64.h
@@ -455,142 +455,121 @@ private:
 };
 
 inline auto
-get_vmcs_field(uint64_t addr, const std::string &name, bool exists)
+get_vmcs_field(uint64_t addr, const char *name, bool exists)
 {
-    auto what = std::string("get_vmcs_field failed: ") + name + " field doesn't exist";
-
     if (!exists)
-        throw std::logic_error(what);
+        throw std::logic_error("get_vmcs_field failed: "_s + name + " field doesn't exist");
 
     return intel_x64::vm::read(addr, name);
 }
 
 inline auto
-get_vmcs_field_if_exists(uint64_t addr, const std::string &name, bool verbose, bool exists)
+get_vmcs_field_if_exists(uint64_t addr, const char *name, bool verbose, bool exists)
 {
-    if (!exists && verbose)
-        bfwarning << "get_vmcs_field_if_exists failed: " << name << " field doesn't exist" << '\n';
-
     if (exists)
         return intel_x64::vm::read(addr, name);
+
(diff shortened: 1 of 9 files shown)


bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -455,142 +455,109 @@ class vmcs_intel_x64
...
 
 inline auto
-get_vmcs_field(uint64_t addr, const std::string &name, bool exists)
+get_vmcs_field(uint64_t addr, const char *name, bool exists)
 {
-    auto what = std::string("get_vmcs_field failed: ") + name + " field doesn't exist";
-
     if (!exists)
-        throw std::logic_error(what);
+        throw std::logic_error("get failed: field doesn't exist");
```
rianquinn: update this to include the name. It's safe to use the "+" here as the throw should never happen. The old code was creating the "what" all the time, which is why it was an issue. The normal case needs to be as fast as possible. The error case can be slow.

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -455,142 +455,109 @@ class vmcs_intel_x64
...
+get_vmcs_field_if_exists(uint64_t addr, const char *name, bool verbose, bool exists)
 {
-    if (!exists && verbose)
-        bfwarning << "get_vmcs_field_if_exists failed: " << name << " field doesn't exist" << '\n';
-
     if (exists)
         return intel_x64::vm::read(addr, name);
 
+    if (!exists && verbose)
+        bfwarning << "get failed: " << name << " doesn't exist" << '\n';
```
rianquinn: same thing here. feel free to provide the function name. slow is fine on the error case

bfvmm/include/vmcs/vmcs_intel_x64.h
```diff
@@ -455,142 +455,109 @@ class vmcs_intel_x64
...
 
 template <class T> void
-set_vmcs_field(T val, uint64_t addr, const std::string &name, bool exists)
+set_vmcs_field(T val, uint64_t addr, const char *name, bool exists)
 {
-    auto what = std::string("set_vmcs_field failed: ") + name + " field doesn't exist";
-
     if (!exists)
-        throw std::logic_error(what);
+        throw std::logic_error("set failed: field doesn't exist");
```
rianquinn: same thing

connojd:
@rianquinn I made the changes last night and forgot to mention you

PR #316, opened 2016-11-16
Make 32-bit read-only data field namespaces

connojd:
This is a big one.  The bulk of the changes are from the vm_exit_instruction_information namespaces


rianquinn:
LGTM

PR #322, opened 2016-11-17
Make 64-bit control field namespaces

connojd:


rianquinn:
this one has issues with coveralls

connojd:
@rianquinn it should be ready now

bfvmm/src/vmcs/test/test.cpp
```diff
@@ -118,6 +118,95 @@ enable_entry_ctl(uint64_t control)
     vmcs::vm_entry_controls::set(ctls | control);
 }
 
+
+void
+proc_ctl_allow1(uint64_t mask)
+{ g_msrs[msrs::ia32_vmx_true_procbased_ctls::addr] |= mask << 32; }
+
+//void
```
rianquinn: Please remove these comments once all of this is done

connojd: I will..those were part of the reason coveralls wasn't passing

PR #328, opened 2016-11-18
Make 64-bit guest-state field namespaces

connojd:


connojd:
^^appveyor finished before travis!!

connojd:
I've thought about the the memory_type in the pat namespace..I noticed after I pushed that the memory_type is already in the x64 namespace and I like it better there rather than listing them in the pa0, pa1, etc.

I was thinking of changing the pa0, pa1 ... namespaces to something like this
```cpp
namespace pa0
{
    [mask]
    [from]
    [name]
    [getters,setters]

    template<class T, class = typename <std::enable_if<foo>>
    auto __is_not_reserved(T mem_type)
    {
        switch (mem_type)
        {
            case x64::memory_type::uncacheable:
            case x64::memory_type::write_combining:
            case x64::memory_type::write_through:
            case x64::memory_type::write_protected:
            case x64::memory_type::write_back:
            case x64::memory_type::uncached:
                return true;

            default:
                return false;
        }
    }

    auto is_not_reserved()
    { /* call __is_not_reserved or throw if field doesn't exist */ }

    auto is_not_reserved_if_exists()
    { /* call __is_not_reserved or warn if doesn't exist */}
}
```
That would allow us to move the `check_pat` functionality into the pat namespace

rianquinn:
I'd leave the is_not_reserved functions out for now. The logic that is in the VMCS checks should work for now.

rianquinn:
and yeah.. the memory type is something that is defined throughout the Intel manual, so there is no need to define it in the VMCS.

connojd:
@rianquinn ready for review

PR #334, opened 2016-11-18
Make 64-bit host-state field namespaces

connojd:


connojd:
@rianquinn ready to merge

PR #338, opened 2016-11-20
Implement vmcs debug printing

connojd:
The implementation here is pretty ugly but it will work for RC 1.1.  It should be revisited eventually.  Placing a dump function in each field's namespace is probably the cleanest way and makes the most sense, unfortunately I don't think I have time to do that before RC 1.1.


bfvmm/src/vmcs/src/vmcs_intel_x64_debug.cpp
```diff
@@ -22,524 +23,716 @@
...
     bfdebug << "Natural Width Control Fields:" << bfendl;
-    PRINT_FIELD(true, VMCS_CR0_GUEST_HOST_MASK);
-    PRINT_FIELD(true, VMCS_CR4_GUEST_HOST_MASK);
-    PRINT_FIELD(true, VMCS_CR0_READ_SHADOW);
-    PRINT_FIELD(true, VMCS_CR4_READ_SHADOW);
-    PRINT_FIELD(true, VMCS_CR3_TARGET_VALUE_0);
-    PRINT_FIELD(true, VMCS_CR3_TARGET_VALUE_1);
-    PRINT_FIELD(true, VMCS_CR3_TARGET_VALUE_2);
-    PRINT_FIELD(true, VMCS_CR3_TARGET_VALUE_31);
+    dump_vmcs_field(cr0_guest_host_mask::addr, cr0_guest_host_mask::name, cr0_guest_host_mask::exists());
```
rianquinn: I like how you added newlines to the other fields since they were really large. I would keep the same thing for all of the fields, so this would be

```cpp
dump_vmcs_field(cr0_guest_host_mask::addr,
                              cr0_guest_host_mask::name, 
                              cr0_guest_host_mask::exists());
```

bfvmm/src/vmcs/src/vmcs_intel_x64_debug.cpp
```diff
@@ -22,524 +23,716 @@
...
 
 void
-vmcs_intel_x64::print_pin_based_vm_execution_controls()
+debug::dump_pin_based_vm_execution_controls()
 {
-    auto controls =
-        vmread(VMCS_PIN_BASED_VM_EXECUTION_CONTROLS);
-
     bfdebug << "----------------------------------------" << bfendl;
     bfdebug << "- Pin VM Execution Controls            -" << bfendl;
```
rianquinn: At a minimum, I would get rid of these blocks and make the output look more like:
https://github.com/Bareflank/hypervisor/blob/master/bfvmm/include/intrinsics/msrs_intel_x64.h#L836

Like I did in this code, if it's 0/1, I called it a "flag" and only print the ones that are enabled, and for the fields, I printed a title, and then "    - <name> = view_as_pointer(<value>)". 

This way as least everything is consistent.

bfvmm/src/vmcs/src/vmcs_intel_x64_debug.cpp
```diff
@@ -22,524 +23,716 @@
...
-    PRINT_CONTROL(VM_EXEC_PIN_BASED_NMI_EXITING);
-    PRINT_CONTROL(VM_EXEC_PIN_BASED_VIRTUAL_NMIS);
-    PRINT_CONTROL(VM_EXEC_PIN_BASED_ACTIVATE_VMX_PREEMPTION_TIMER);
-    PRINT_CONTROL(VM_EXEC_PIN_BASED_PROCESS_POSTED_INTERRUPTS);
+    dump_vm_control(pin_based_vm_execution_controls::external_interrupt_exiting::name,
+                    pin_based_vm_execution_controls::external_interrupt_exiting::is_enabled(),
+                    ia32_vmx_true_pinbased_ctls::external_interrupt_exiting::is_allowed0(),
+                    ia32_vmx_true_pinbased_ctls::external_interrupt_exiting::is_allowed1());
+
+    dump_vm_control(pin_based_vm_execution_controls::nmi_exiting::name,
```
rianquinn: Please move the implementation to the headers as inline functions and ditch the .cpp

bfvmm/src/vmcs/src/vmcs_intel_x64_debug.cpp
```diff
@@ -22,524 +23,716 @@
...
 }
 
-#endif
+void
+debug::dump_vmcs_field(vmcs::field_type addr, const char *name, bool exists)
+{
+    bfdebug << name << ": " ;
+
+    if (!exists)
+        bfinfo << "doesn't exist" << bfendl;
```
rianquinn: I like this, just format it like I stated above. But yeah... I like that you replace <value> with "doesn't exist".

bfvmm/src/vmcs/src/vmcs_intel_x64_debug.cpp
```diff
@@ -22,524 +23,716 @@
...
+}
+
+void
+debug::dump_vm_control(const char *name, bool is_set,
+                       bool is_allowed0, bool is_allowed1)
+{
+    bfdebug << name << ": ";
+
+    bfinfo << "value == " << is_set
+           << ", is_allowed0 == " << is_allowed0
```
rianquinn: To simplify things for now... ditch the is_allowed stuff... and just format this like the flags fields I showed above. If the flag is enabled.... show it... otherwise don't. If they want to know what is allowed / not allowed, the current MSR dump code shows that.

rianquinn:
@connojd coveralls is failing. And it's an annoying one as it doesn't know what is not being covered. I've seen it before.... not fun to track down.

connojd:
I think it was the true branch in the secondary controls function that wasn't getting hit.  I added some units tests for the dumpers so it shouldn't be a problem in the future.

rianquinn:
@connojd were there unit tests missing for that?

connojd:
Yeah there weren't any for the debug functions

Issue #336, opened 2016-11-19
[RFC]: Deprecating GCC 5.x


rianquinn:
I think it's time to deprecate GCC 5.x (for the cross compiler only, native compilation will still support GCC 5.x):

- Currently we are testing a lot of compilers, and it's really not needed. Getting rid of GCC 5.x will reduce maintenance and testing time. 
- GCC 6.x doesn't seem to be able to compile GCC 5.x, so the local compilers don't work on Ubuntu 16.10 unless you specify GCC 6.x only (Ubuntu 16.10 updated to GCC 6.x)
- GCC 5.x seems to have bugs that cause crashes in libcxx. Using GCC 6.x or Clang resolves these bugs. 

In addition to these problems, we don't have Clang specific cross compilers, and instead we have GCC 5.2 which includes Clang 3.8 and GCC 6.2 which includes Clang 3.9. Instead, I would like to decouple newlib from GCC, which would allow us to have a Clang only cross compiler and a GCC only cross compiler. 

Supported cross compilers would include:
- GCC 6.1
- GCC 6.2
- Clang 3.8
- Clang 3.9

Thoughts?

At the moment this is a blocker for completing the libcxx unit tests as GCC 5.2 is crashing them (as stated above), so likely I will start a PR for this ASAP so if you have any concerns, speak up 😄

brendank310:
I think it'd be easier to comment on based on how it affects platform default compilers than the compiler versions themselves.

rianquinn:
Ok... so I have a Clang only build, but as I started to break this down I noticed that newlib is not being compiled properly at all, and once I started to try, I noticed that it's really not happy about how we are not compiling with a target platform. newlib REALLY badly wants a target platform (i.e. x86_64_something_elf instead of x86_64_elf). There are some #defines that you can fudge, but then libcxx gets mad, so you end up with this game with no end of someone not being happy... newlib or libcxx. 

Short story is... our own libc is probably the best solution here since we need one for the hyperkernel, and have had a custom libc on the list for a while now, so I'm going to look into it to see how bad it's going to be. I'll report back once I have more info.

PR #341, opened 2016-11-22
Deprecate GCC 5

rianquinn:
This patch set removes GCC 5 and defaults to Clang 3.8 as the
cross compiler. In addition, newlib has been updated, and is now
linked in as a shared library instead of a static library.
Finally, when Clang is used, only Clang is used, unlike
before were GCC was used to compile Newlib

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/336


rianquinn:
Merging. Windows will be corrected once we get the proper compilers up on the server

PR #355, opened 2016-11-26
Implement remaining vmcs unit tests

connojd:
This commit implements unit tests for the vmcs
state, vmm_state, and host_vm_state classes.


.travis.yml
```diff
@@ -63,7 +63,7 @@ matrix:
...
+      - coveralls --build-root makefiles -e build_libbfc -e build_libcxx -e build_libcxxabi -e build_newlib -e source_libbfc -e source_libcxx -e source_libcxxabi -e source_llvm -e source_newlib -e sysroot -e bfdrivers/src/arch -e bfm/src/arch -e include/hippomocks.h -e include/json.h -e include/unittest.h -e include/new_delete.h -e include/gsl -e include/exception.h -e bfunwind -e bfvmm/src/intrinsics -e bfvmm/src/vcpu_factory/src/vcpu_factory.cpp -e bfvmm/src/vmcs/src/vmcs_intel_x64_promote.asm -e bfvmm/src/vmcs/src/vmcs_intel_x64_resume.asm -e bfvmm/include/vmcs/vmcs_intel_x64_promote.h -e bfvmm/include/vmcs/vmcs_intel_x64_resume.h --gcov-options '\-lp'
```
rianquinn: remove the rest of the VMCS files. The "asm" files can be flat out removed because they are not being checked by gcov, and the header files should also be removed. If you end up with an issue with the header files, just provide a dummy implementation `void vmcs_promote(){}` etc... and call them in the unit test.

connojd:
@rianquinn all the vmcs files are gone now

PR #361, opened 2016-11-28
Exit Qualification Optimizations

rianquinn:
The exit qualification fields need to be parsed as fast as
possible, and reading the VMCS on each sub-field is not
a good idea, so this patch provides the ability to
read the sub-fields once the fields have been read.


rianquinn:
@connojd / @brendank310 passed, ready to merge

PR #362, opened 2016-11-29
Add Cache Flush

rianquinn:
This adds the cache flush intrinsic, and adds cache flush to
unique_map. Altough it was working before, it seems sensible
to flush the cache on an unmap to ensure all mapped memory
is written.

### DIFF
--- a/bfvmm/include/intrinsics/cache_x64.h
+++ b/bfvmm/include/intrinsics/cache_x64.h
@@ -24,6 +24,7 @@
 
 extern "C" void __invd(void) noexcept;
 extern "C" void __wbinvd(void) noexcept;
+extern "C" void __clflush(void *addr) noexcept;
 
 // *INDENT-OFF*
 
@@ -31,11 +32,20 @@ namespace x64
 {
 namespace cache
 {
+    using pointer = void *;
+    using integer_pointer = uintptr_t;
+
     inline void invd() noexcept
     { __invd(); }
 
     inline void wbinvd() noexcept
     { __wbinvd(); }
+
+    inline void clflush(integer_pointer addr) noexcept
+    { __clflush(reinterpret_cast<pointer>(addr)); }
+
+    inline void clflush(pointer addr) noexcept
+    { __clflush(addr); }
 }
 }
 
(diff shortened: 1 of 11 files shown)


rianquinn:
@josephsharkey / @ktemkin / @brendank310 / @connojd 

I added the clflush to the unique_map. I was unable to find in the Intel SDM how multiple virt to phys mappings on the same core are handled with respect to write-back memory. Seems to me that if you happen to map the same phys address to two different virtual addresses that you would end up with a cache coherency issue because they both have their own cache, in which case one would overwrite the other. Correct me if I am wrong

So with that said, I'm puzzled why they don't seem to address this issue with respect to VMX in the manual (maybe I just cannot find the section). Without VMX.. mapping the same physical memory twice seems like a terrible idea, but it will almost certainly happen with respect to VMX. For example, the host mapping guest memory for emulation, or a guest mapping another guest's memory for comms. I guess the question is... how does the hardware deal with this? So far it just seems to work, but I'd like to know why, and I cannot find the text in the manual that addresses this issue.

rianquinn:
I did a lot more reading on this called "aliasing"... (bringing back horrible memories from Comp Arch 101), so long as the page tables map memory the same (i.e. the aliases are all WB) the CPU takes care of the issue. I was able to find a reference as well that stated that for Intel, the physical address is used for the cache and not the virtual which eliminates the issue as well.

rianquinn:
@brendank310 Ok, based on what we learned, I have left the cache_flush code in there, but it's not automatically run on unmap as its likely not needed. Future versions of the map code should lookup the cache properties of the map prior to mapping so that mapping is consistent but for now, we should be able to get away with WB only. Once the tests complete, this should be good to merge.

rianquinn:
@brendank310 After sleeping on it, I'm still not happy with this solution, and I think in general there might be a larger issue with how we are programming the PATs (i.e. I think the hypervisor is currently set to uncached memory for everything), so I'm working on a more comprehensive patch that will use the PATs properly, and allow unique_map to setup mappings the same as the guest's mappings, or at least yell when it can't. This should be a lot safer.

PR #365, opened 2016-11-30
Cache Flush / PAT Mapping / Bug Fixes

rianquinn:
This PR adds cache flush to the repo, it also adds support for
the PATs. Up until this point, the hypevisor has been executing
with uncached memory, now it uses WB memory, and the maps can
be setup to use any memory type you want. If you map in guest
memory, it will match the guest mappings to prevent
inconsistent aliases.

This patch also removed a couple of bugs with the vCPU, and
Hippomocks, and added a mcro to silence verbosity with VMWare.


bfvmm/include/intrinsics/x64.h
```diff
@@ -33,105 +32,104 @@
 
 namespace x64
 {
-    constexpr const auto page_size                               = 0x1000UL;
-    constexpr const auto page_shift                              = 12UL;
-    constexpr const auto cache_line_size                         = 64UL;
-    constexpr const auto cache_line_shift                        = 6UL;
+    constexpr const auto page_size                                   = 0x1000UL;
+    constexpr const auto page_shift                                  = 12UL;
```
connojd: Was the whitespace intentional here?

PR #381, opened 2016-12-09
Refactored argument filtering in bareflank_gcc_wrapper

ilovepi:
The bareflank_gcc_wrapper's argument parsing reordered the CLI options
passed to it. This caused issue if there was an ordering dependency in
the compile or link options, and also prevented positional compiler
options from being correctly parsed. These changes allow:

- consistent argument ordering
- provide handling for positional arguments
- correct handling of '-Wl,' options
- adds support for '-mllvm', '-Xclang', '-Xlinker',
    and '-Wl,--plugin-opt' CLI options

By removing the COMMON_ARGS environment variable, and limiting the
options to only two lists, the ordering of arguments can be preserved.
Additionally, I found that no '-Wl,*' options were ever passed to the
linker. Earlier in the script there was a match using '-W*', which
essentially prevented these linker directives from being handled.

I also added argument processing for '-mllvm', '-Xclang', '-Xlinker',
and '-Wl,--plugin-opt' CLI options. Some of these are notably Clang
specific, or require the gold linker with plugin support. 'Xlinker' is
supported by both GCC and Clang, as a way to directly pass arguments to
the linker.

Signed-off-by: Rian Quinn <rianquinn@gmail.com>

### DIFF
--- a/tools/scripts/bareflank_gcc_wrapper.sh
+++ b/tools/scripts/bareflank_gcc_wrapper.sh
@@ -154,20 +154,44 @@ done
 # Filter Arguments
 # ------------------------------------------------------------------------------
 
-COMMON_ARGS_INDEX=0
 COMPILE_ARGS_INDEX=0
 LINK_ARGS_INDEX=0
 SOURCE_ARGS_INDEX=0
 OBJECT_FILE_ARGS_INDEX=0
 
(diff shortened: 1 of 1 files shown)


rianquinn:
This looks great. Thanks a ton for the PR.

PR #396, opened 2016-12-20
Arch Linux setup script

JohnnyLeone:
I've tested it with the docker and local compilers.
Tests, driver loads and unloads are fully functional.

Pacman does not provide a "yes" flag to skip the install requests. There is manual workaround for it, but you have to pipe the yes command in pacman install command. - FIXED
@connojd  gave me a hint to use the --noconfirm ,thanks.

Hopefully someone find it useful.

### DIFF
--- a/tools/scripts/setup_archlinux.sh
+++ b/tools/scripts/setup_archlinux.sh
@@ -42,24 +42,23 @@ parse_arguments $@
 
 install_pacman_tools() {
     sudo pacman -Syu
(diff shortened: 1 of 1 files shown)


rianquinn:
LGTM

@JohnnyLeone If possible, I would like to wait to merge this until after we release 1.1, which should be in a in about 3 weeks (need time to finish testing). Once 1.1 is released, we will begin development of 1.2 and this will be the first PR to be merged as it looks great. 

Thanks a ton

PR #397, opened 2016-12-22
CR3 / Paging Update

rianquinn:
This patch makes several changes to the CR3 / paging logic:
- Memory usage is about as small as it's going to get. With this patch,
  all that is needed is the page table itself, and a list of the next
  set of page tables. Mapping / unmapping performance should be about
  the same, if changed at all
- 1g, 2m and 4k granularity has been added
- You can now swap between different granularities without corrupting
  the page tables.
- A new capacity function has been added to measure how much capacity
  has been allocated vs the total size. This is mainly used by the
  unit tests, but you can use it in your code if your curious about
  how much memory your consuming.
- virt_to_pte has been added so that you can locate a PTE for a
  given virtual address.

### DIFF
--- a/bfvmm/include/memory_manager/map_ptr_x64.h
+++ b/bfvmm/include/memory_manager/map_ptr_x64.h
@@ -51,6 +51,40 @@ namespace bfn
 template <class T>
 class unique_map_ptr_x64;
 
+/// Lower
+///
+/// @param ptr the pointer to mask
+/// @return the lower 12 bits of ptr
+///
+inline auto lower(uintptr_t ptr) noexcept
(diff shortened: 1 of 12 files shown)


rianquinn:
I alos cleaned up a bunch of bugs with this PR as well. Specifically:
- Using the CR3 map with an unaligned address no longer leaks memory when unmapped. 
- Mapping code is no longer using memory after it has been unmapped (should have been causing triple faults)
- Mapping in a page no longer accepts an offset which didn't make sense
- Mapping in a page with an offset no longer maps in an extra page if it's not needed.

PR #401, opened 2016-12-31
VMLauch Mods

rianquinn:
This patch adds a traditional vmlaunch which is needed to start
guests with the VMCS. Now, you can declare a guest state as a
guest and not a host (the default), and vmlauch will be executed
without the demotion.

To support this, the vCPU has been relaxed so that it can be
not only be started, but it can also be resumed (i.e. run twice).
We also provided RIP / RSP fields in the state class so that
these registers can be defined for guest support.

Finally the VMCall completetion code has been moved so that
a vmcall can be completed from a different VM Exit.


bfvmm/include/vmcs/vmcs_intel_x64_state.h
```diff
@@ -277,6 +279,19 @@ class vmcs_intel_x64_state
...
+    { return 0; }
+    virtual integer_pointer rsp() const
+    { return 0; }
+
+    virtual void set_rip(integer_pointer val) const
+    { (void) val; }
+    virtual void set_rsp(integer_pointer val) const
+    { (void) val; }
+
+    virtual bool is_guest()
```
brendank310: Is this a useful bool for vmcs state? Isn't being associated with the VMCS implying it is a guest? I think I know what your getting at (whether it was demoted or launched) but it would be nice if we could give them similar treatment.

bfvmm/src/memory_manager/src/page_table_entry_x64.cpp
```diff
@@ -94,14 +94,22 @@ page_table_entry_x64::set_ps(bool enabled) noexcept
...
+page_table_entry_x64::set_pat_4k(bool enabled) noexcept
 { *m_pte = enabled ? set_bit(*m_pte, 7) : clear_bit(*m_pte, 7); }
 
 bool
+page_table_entry_x64::pat_large() const noexcept
+{ return is_bit_set(*m_pte, 12); }
+
+void
+page_table_entry_x64::set_pat_large(bool enabled) noexcept
+{ *m_pte = enabled ? set_bit(*m_pte, 12) : clear_bit(*m_pte, 12); }
```
brendank310: Rodney has got to see this.

bfvmm/include/vmcs/vmcs_intel_x64_state.h
rianquinn: Basically, the state_host and state_vmm leave this as it's default which is false. As a result, a vmlaunch will demote. The state_guest sets this to true, and as such, all vmlaunches use a standard vmlaunch (i.e. no return, no demote). 

So question is... based on that explanation, what change are you looking for?

bfvmm/src/memory_manager/src/page_table_entry_x64.cpp
rianquinn: the use of a ternary? 😄

bfvmm/include/vmcs/vmcs_intel_x64_state.h
brendank310: No change at the moment, I think we should strive for the APIs to be orthogonal to the collection of cases is all. Just something to keep in mind.

rianquinn: agreed.

bfvmm/src/memory_manager/src/page_table_entry_x64.cpp
brendank310: Yes the ternary :1st_place_medal:

rianquinn: lol. @5n4k3

5n4k3: Lol. Brendan, when I saw your "hell has frozen over" email, I was expecting to see functors, but this is almost as good.

PR #404, opened 2017-01-10
Fixed bf-gcc bug

RKX1209:
Some header files are using #include_next extension, for example, cmath includes math.h.
It searches include directories in the same order as that specified by -isystem option.
So system include directories must be reordered correctly.

-isystem on a system include directory causes errors
http://stackoverflow.com/questions/37218953/isystem-on-a-system-include-directory-causes-errors


rianquinn:
This patch makes a lot of sense. Thanks a ton for the contribution. Let use know if you need anything else, and feel free to toss in feature requests if Bareflank is missing anything else you need.

PR #416, opened 2017-01-25
VM Launch Mods #2

rianquinn:
This patch provides a bunch of mods needed to launch virtual
machines. These include:
- The ability to setup more than one sysroot, which can be
  used by an extension to create it's own sysroots.
- A complete re-write of the ELF loader to simplify it's code
  base and provides support for additional binary types including
  binaries with and without PIE, as well as shared libraries.
  The ELF loader also includes code to get various different
  properties associated with the ELF file including dependencies,
  permission info, and how it was compiled
- the "misc" module is now a binary that includes "__dso_handle". 
  This binary can be used to statically link binary only libraries needed
  by the compiler. This also removes the need to disable __cxa_atexit.
- various different bug fixes

Note: This patch offically drops support for GCC in favor of
  Clang

@ilovepi


brendank310:
LGTM, ready to merge once passes CI.

rianquinn:
@brendank310 👍

PR #428, opened 2017-02-14
Added VM test script

JWZepf:
Added the script to be put into each VM to test Bareflank. Script located in hypervisor/tools/tests/BFTest.sh

### DIFF
--- a/tools/tests/BFTest.sh
+++ b/tools/tests/BFTest.sh
@@ -1,0 +1,89 @@
+check_status()
+{
+    if [ $? -ne 0 ] ; then
+        echo BFTest FAILED > /dev/ttyS1
+        sleep 31
+    fi
+}
+
+test_hypervisor()
+{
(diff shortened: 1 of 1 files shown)


tools/tests/BFTest.sh
```diff
@@ -0,0 +1,89 @@
+check_status()
```
rianquinn: Doesn't look like this function is used. Also, you need to put the copyright header in place, and make sure thta is starts with `#!/bin/bash -e`

JWZepf: check_status is used after every command. It essentially does what the -e flag on the bash shebang does, but it prints out a statement to the serial log file that the Jenkins script is looking for. If I just did the -e flag, the Jenkins script would stall out if there was an error in a command. I added the copyright header and the shebang to the top of the script.

rianquinn:
Looks great. I will merge once the tests pass

PR #432, opened 2017-02-27
Updated setup_cygwin.sh

JWZepf:
Made a couple changes to make setup work on cygwin:

- Line 43: Added curl and wget to packages installed with cygwin setup. I imagine most systems already have this installed, but since it is needed later in the script it would make sense to get it at this stage just in case. The bare minimum cygwin installation does not include curl and wget.
- Line 48: The output file name has to follow the -o flag. Currently, it follows the -L flag which confuses the terminal when figuring out where you want the output to go.
- Line 50: pushd gets a "too many arguments" error when using `cmake-*`. Since cmake-3.6.2 is hardcoded 2 lines up, I don't think its a problem to hardcode it here and that fixes the issue.
- Line 75: The latest cygwin version is 2.7.* and this is a pretty new release (I think) so this script might not have been updated to support the newest cygwin client.

### DIFF
--- a/tools/scripts/setup_cygwin.sh
+++ b/tools/scripts/setup_cygwin.sh
@@ -40,14 +40,14 @@ parse_arguments $@
 # ------------------------------------------------------------------------------
 
 install_common_packages() {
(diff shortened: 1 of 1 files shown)


rianquinn:
@JWZepf If you get a chance, feel free to see if you can figure out how to get AppVeyor working again with these changes. You might need to install cygwin into your own directory instead of using the one that AppVeyor comes with.

PR #434, opened 2017-03-03
Remove Cross Compilers / ELF Loader Update

rianquinn:
This patch removes the need for the cross compilers. You can now
install clang using your distro's package manager, and the scripts
will tell clang to cross compile, linking with the linker of your
choice, or defaulting to "ld". Note that the default clang that
is expected is clang 3.8, but you can set the compiler to whatever
you want, including GCC if you create the cross compiler youself.

The ELF loader was not working with ld.gold or ld.lld because these
linkers craete a rela.plt and use the old .ctors / .dtors sections.
They also do not properly label the eh_frame section. This patch
uses an updated ELF loader that is actually slated for 1.2, but has
these bug fixes in it.

Signed-off-by: Rian Quinn <“rianquinn@gmail.com”>


rianquinn:
@brendank310 Feel free to give this one a look. A lot of files were touched, but this really was just replacing the ELF loader with the one that I am working on for 1.2, and getting rid of the cross compiler. While doing that, I was able to clean up AppVeyor, and it's building the VMM now during the test, something it has not done for some time now. @ilovepi feel free to give this a shot. I have tested this with ld, ld.gold and ld.lld on Linux, and x86_64-elf-ld.exe on Windows. Sadly Windows needs a custom version of LD since the one that cygwin comes with does not support a sysroot and only links PE/COFF.

PR #435, opened 2017-03-08
bf_test.sh: Cygwin Update

JWZepf:
Cygwin doesn't do unit testing, so it shouldn't run `make test`. This extension checks to see if the OS running the script is Cygwin before running `make test`.


JWZepf:
Changed READMEs to provide instruction with the Fedora fixes

PR #441, opened 2017-03-28
Add support for bareflank on kvm

connojd:
There were two problems encountered when trying to run
Bareflank as a guest on KVM:

    1. KVM doesn't pass through access to the IA32_PERF_GLOBAL_CTRL
    msr (address 0x38F).

    2. KVM doesn't set the lock bit in the IA32_FEATURE_CONTROL msr
    (address 0x3A).

This commit is a minimal change to prevent unchecked accesses to
IA32_PERF_GLOBAL_CTRL and to allow for VMXON to succeed by setting
the enable_vmx_outside_smx bit (bit 2) and then the lock bit (bit 0)
in IA32_FEATURE_CONTROL.

Note that this patch was tested with Ubuntu 16.10 and Arch Linux
guest VMs running on an Ubuntu 16.10 KVM.

### DIFF
--- a/bfvmm/include/intrinsics/cpuid_x64.h
+++ b/bfvmm/include/intrinsics/cpuid_x64.h
@@ -1031,6 +1031,150 @@ namespace cpuid
         }
     }
 
+    namespace arch_perf_monitoring
+    {
+        constexpr const auto addr = 0x0000000AUL;
+        constexpr const auto name = "arch_perf_monitoring";
+
+        namespace eax
+        {
+            namespace version_id
+            {
+                constexpr const auto mask = 0x000000FFUL;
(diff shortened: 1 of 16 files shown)


connojd:
@rianquinn codacy says theres a function not being executed but coveralls says otherwise. I guess I'm missing something?

rianquinn:
Ignore codacy for the hypervisor. I will turn that off

rianquinn:
LGTM. 

@connojd We might want to switch to xxx_if_exists for a lot of those calls, but this is great for now.

connojd:
@rianquinn yes thats probably a good idea for future nesting attempts

PR #442, opened 2017-03-29
Remove if_exists in exit handler

connojd:
This removes get_if_exists and set_if_exists from the rdmsr and wrmsr handlers


rianquinn:
LGTM

PR #453, opened 2017-05-28
Fix for build_windows.bat

### LINKED ISSUE #451: [Bug] Fix for the build_windows.bat file
Hey,

I noticed that when you try to execute `make driver_load` on a different drive than `C` on Windows, it complains about `LaunchBuildEnv.cmd` not being a valid command/file.
To fix this, I had to change the `cd` commands to include `/D`, which forces a drive change.
From the `cd` documentation:
```
Use the /D switch to change current drive in addition to changing current
directory for a drive.
```

In short:
`cd C:\ewdk\` -> `cd /D C:\ewdk\`.
`cd %1` -> `cd /D %1`

Randshot:
Fixes #451

### DIFF
--- a/tools/scripts/build_windows.bat
+++ b/tools/scripts/build_windows.bat
@@ -1,7 +1,7 @@
 @echo off
 
-cd C:\ewdk\
+cd /D C:\ewdk\
 call LaunchBuildEnv.cmd
 
-cd %1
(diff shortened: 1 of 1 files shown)


rianquinn:
LGTM, I will merge once the tests pass

Issue #461, opened 2017-06-23
[Issue] bfm: ioctl failed: `IOCTL_START_VMM`


Randshot:
Hi,

two of my friends seem to have issues with starting the hypervisor.
It loads (`make load`) just fine, but when they try to start (`make start`) it, the get an error:
```
bfm: ioctl failed: `IOCTL_START_VMM`
```

First I though that it was just a single case, but now another friend of mine encountered the same issue.
For me it works just fine, so I don't get why it won't work for them.
Both of them use Windows 10.

rianquinn:
Check via serial to see what message came up. Likely it's something simple like virtualization extensions are not turned on.

Randshot:
Serial?

rianquinn:
Bareflank supports serial output in addition to `make dump`

Randshot:
Eh, how do I do that?
I tried PortMon, but that didn't let me capture anything.

rianquinn:
You need to identify which port your connected to and then tell Bareflank to use that port:
http://bareflank.github.io/hypervisor/html/index.html#serial

nick-kvmhv:
My situation it must be something related to Windows 10 itself: I have no problem starting it on Ubuntu on the same laptop. Windows 10 gives me identical error when I boot in in KVM or as the main OS. If there is more output in serial than in DebugView, I'll try to dust off my old serial over USB cable and check.

rianquinn:
Yeah, you should get more input as it died safely which means that it knows what the issue is, you just need to get access to the error. Make dump only works when the vCPUs are up and running.

nick-kvmhv:
Before I figure out how to connect that serial cable to a laptop with no serial ports, here is what I get in DbgView when I try to start it:
[BAREFLANK ERROR]: start_vmm failed
[BAREFLANK ERROR]: IOCTL_START_VMM: ioctl_start_vmm failed: 8000000000000030 - ENTRY_ERROR_VMM_START_FAILED

rianquinn:
yeah that just means that it attempted to start but failed. Once option would be to create a global debug ring and only dump from the debug ring, in which case you could get the output from make dump without needing serial

nick-kvmhv:
When running it under KVM, I got this in COM1 output when starting:
----------------------------------------
- Standard Exception Caught            -
----------------------------------------
vector

rianquinn:
[...]

PR #462, opened 2017-06-24
Fix for last gdt descriptor not getting copied/allocated

nick-kvmhv:
The fix that resolved the issue we discussed here: https://github.com/Bareflank/hypervisor/issues/461
The rationale behind this change is that m_gdt_reg.limit is coming directly from SGDT and therefore is size minus one.

### DIFF
--- a/bfvmm/include/intrinsics/gdt_x64.h
+++ b/bfvmm/include/intrinsics/gdt_x64.h
@@ -219,7 +219,7 @@ public:
             m_gdt_reg.base = x64::gdt::base::get();
             m_gdt_reg.limit = x64::gdt::limit::get();
 
-            std::copy_n(m_gdt_reg.base, (m_gdt_reg.limit+1) >> 3, std::back_inserter(m_gdt));
+            std::copy_n(m_gdt_reg.base, (m_gdt_reg.limit + 1) >> 3, std::back_inserter(m_gdt));
         });
     }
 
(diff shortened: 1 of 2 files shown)


rianquinn:
After having a chance to think about this more, I'm confused as to why this is needed. Before I merge, I want to see if I can reproduce the same issue so that I can study the root cause (which might just be my ignorance). 

Also, astyle is made because it should be ` + 1` and not `+1`.

nick-kvmhv:
On why it is needed, I think it is because as OSWiky states here http://wiki.osdev.org/Global_Descriptor_Table
```The offset is the linear address of the table itself, which means that paging applies. The size is the size of the table subtracted by 1. This is because the maximum value of size is 65535, while the GDT can be up to 65536 bytes (a maximum of 8192 entries). Further no GDT can have a size of 0.```

rianquinn:
Huh, I knew about the -1 for the IDT but somehow missed that for the GDT. This being the case, this line should have a -1 similar to the IDT
https://github.com/Bareflank/hypervisor/blob/master/bfvmm/include/intrinsics/gdt_x64.h#L242

The IDT code can be seen here:
https://github.com/Bareflank/hypervisor/blob/master/bfvmm/include/intrinsics/idt_x64.h#L194

This will likely need a small fix to the unit tests as well. Can you make these changes as part of your PR, as well as fix the astyle error?

nick-kvmhv:
Not sure if I figured out the style issues, but I added the created GDT correction and also updated the unit tests.

rianquinn:
LGTM +1

PR #464, opened 2017-07-12
fix vmcall registers documentation

Alpha-10000:


rianquinn:
Nice catch, thanks for the PR. If possible, please provide the same PR here (our v1.2 code):
https://github.com/Bareflank/bfsdk/blob/master/include/bfvmcallinterface.h#L289

PR #470, opened 2017-08-16
Merge Repos into the Hypervisor Repo

rianquinn:
This PR mereges all of the repos into a single repo for simplificty. It
also adds static / shared library build types, release / debug builds,
and ensures parallel builds work properly.

Signed-off-by: Rian Quinn <“rianquinn@gmail.com”>


rianquinn:
Ok, this is looking better. Still a lot of cleanup to do, which I will be working on next.

PR #473, opened 2017-08-19
Msbuild updates

rianquinn:


rianquinn:
Ok, the hypervisor repo is building on Windows with both VS and Cygwin, with both Static and Shared libraries. Getting close to having full Windows support again.

PR #475, opened 2017-08-22
BFM cleanup

rianquinn:
This PR completes the BFM work as part of the build system update.


rianquinn:
this PR fixed both Astyle and Clang Tidy which were not working since the merge of the repos.

PR #487, opened 2017-09-13
Added unit tests for new intrinsics

JWZepf:
Updated unit tests to support the updated intrinsics


rianquinn:
@JWZepf you need to convert __builtin_memcpy to just a regular memcpy, and it appears that you are missing just a couple of unit tests for the intrinsics.

Issue #498, opened 2017-09-28
[RFC] Build System Support for Aarch64


JaredWright:
# Description
The current Bareflank build system supports building all components and external dependencies for the x86_64 architecture only. In order to expand support to other architectures, numerous changes to the build system must be made. This RFC proposes a list of changes to the build system that must be accomplished immediately in the short term, and also proposes some longer-term changes that will be more difficult or time consuming to implement.

There are a lot of proposed changes in this RFC, and it would probably make sense to break them down into smaller, separate RFCs after feedback is received. I'm hoping the community can use this RFC to come to consensus about necessary changes to the build system as a whole.

# Proposed Short Term Changes

The immediate short-term goal for build system is to allow for Bareflank components and external dependencies to build for both x86_64 and aarch64 by means of a cross-compiler toolchain. The changes proposed in this section are the bare minimum needed to add basic ARM support to the Bareflank build system:

* Add a mechanism to specify a cross-compile toolchain and/or architecture as input to the Bareflank build system. For an immediate first step, update the env.sh script to allow specifying a target OS and architecture as a parameter.
* Remove all checks from the build system that detect which OS and toolchain you have installed, and use the toolchain configured with env.sh instead.
* Remove nasm as a dependency for ARM builds (or remove it all together?). Use the assembler that comes with the clang toolchain
* Create a new toolchain file for aarch64 (hypervisor/bfsdk/cmake/CMakeToolchain_Aarch64) that will update all necessary cmake variables to make the build system use an ARM toolchain.
* Update or (more desirable) get rid of compiler_wrapper.sh in favor of calling the configured toolchain directly

# Proposed Long Term Changes

The changes proposed in this section are more long-term oriented goals that would help simplify and restructure the Bareflank build system, but will likely require more significant effort than the bare-minimum goals listed above:

* Update build system input mechanism so that it is compatible with bitbake/yocto. In other words, make sure that a Yocto-based toolchain can easily configure Bareflank's cmake system to build for the correct OS/architecture in the context of a larger project.
* Remove almost all configuration from the top-level CMakeLists.txt file, and rather use it as a way to orchestrate smaller, more specialized cmake files under hypervisor/bfsdk/cmake.
* Restructure cmake files that download external dependencies (top-level CMakeLists.txt file and hypervisor/bfsdk/CMakeLists.txt). Break these out into one/several separate cmake files under hypervisor/bfsdk/cmake.
* Allow external dependencies to download in parallel
[...]

PR #499, opened 2017-10-06
Build System Support for Aarch64 #498

JaredWright:
Added Vagrant support for the latest three versions of Ubuntu

1) Added a Vagrantfile

2) Added provisioning scripts for each version of Ubuntu

[RFC] Build System Support for Aarch64 #498

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,15 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+    config.vm.provider "virtualbox" do |vb|
+        vb.memory = "2048"
+        vb.cpus = 2
+    end
+
+    config.vm.define "ubuntu17_04", primary: true do |ubuntu17_04|
+        ubuntu17_04.vm.box = "ubuntu/zesty64"
+        ubuntu17_04.vm.provision "shell",
+            path: "scripts/vagrant/provision_ubuntu17_04.sh"
+    end
+end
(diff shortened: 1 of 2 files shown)


JaredWright:
Question for @rianquinn and @connojd : I tested that Bareflank builds successfully in Ubunutu VMs for 16.04, 16.10, and 17.04. @cpavlina brought up a good point that we might want to just provide one ubuntu VM for vagrant so that we don't have to maintain all three. What are your  thoughts? Provide the option to build in different ubuntun ditros? Or just keep it to one version of Ubuntu?

JaredWright:
Also, to use the provided Vagrant file, you can do the following:

```
vagrant up    # defaults to 17.10
vagrant ssh -c "cmake .. && make -j2"
```

Or you can pick a particular version of Ubuntu:
```
vagrant up ubuntu16_10   # picks the 16.10 VM
vagrant ssh ubuntu16_10 -c "cmake .. && make -j2"
```

rianquinn:
I like the idea of having all of the versions we support so that it is easy to show instructions, but also provide scripts testing building. Couple of questions:
- Is there a way to use a different hypervisor other than VirtualBox?
- Does 6.10 need an update of cmake? We only need 3.6

JaredWright:
1) Yes, you can use other hypervisors as the so-called "provider" for vagrant. I know that VMWare fusion is supported, and I sort-of got vagrant working with KVM at some point

2) I believe that Ubuntu 16.10 comes with cmake 3.5.2

rianquinn:
Ok, for now, lets just support 17.04, and 17.10 moving forward. If VMWare Fusion is supported, we should leverage this to do our functional tests on the Mac Mini server we have in-house. 

The only other comment I have is put all of the apt-get install commands onto a single line as it is faster.

JaredWright:
Ok, I removed the provisioners and configs for Ubuntu 16.04 and 16.10, but I do not have VMWare Fusion, so I can't test out changes for that. It should already support VMWare, and you can add settings the same way as there are some currently for VirtualBox

rianquinn:
LGTM. Once the tests are done I will merge.

PR #508, opened 2017-10-10
CMake Global Variable Migration

JaredWright:
Began migrating global cmake variables into a new cmake
configuration file directory and global_vars.cmake. Started removing
redundant or unused cmake variables

[RFC] #499


CMakeLists.txt
```diff
@@ -305,16 +297,16 @@ if(NOT WIN32)
...
         -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
         -DCMAKE_TOOLCHAIN_FILE=${VMM_TOOLCHAIN_FILE}
         -DENABLE_COVERAGE=${ENABLE_COVERAGE}
         -DENABLE_DYNAMIC_ASAN=${ENABLE_DYNAMIC_ASAN}
         -DENABLE_DYNAMIC_USAN=${ENABLE_DYNAMIC_USAN}
         -DENABLE_TIDY=${ENABLE_TIDY}
         -DENABLE_ASTYLE=${ENABLE_ASTYLE}
-        -DBUILD_SHARED_LIBS=${BAREFLANK_HOST_SHARED_LIBS}
-        -DBUILD_STATIC_LIBS=${BAREFLANK_HOST_STATIC_LIBS}
+        -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
```
rianquinn: No way this is going to work. We have "HOST" and "TEST" static / shared variables because unit tests always need to be compiled using static libraries, while the cross compiled components can be static or shared depending on what the user is looking for.

bfsdk/cmake/CMakeGlobal_Project.txt
```diff
@@ -16,147 +16,33 @@
 # License along with this library; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
-# ------------------------------------------------------------------------------
-# Includes
-# ------------------------------------------------------------------------------
-
-if(NOT DISABLE_CTEST)
-    include(CTest)
```
rianquinn: CTest needs to be disabled for the super builds because they do not have unit tests.

bfsdk/cmake/CMakeGlobal_Project.txt
```diff
@@ -16,147 +16,33 @@
...
 endif()
 
-if(NOT CMAKE_BUILD_TYPE)
-    if(ENABLE_UNITTESTING OR ENABLE_DEVELOPER_MODE)
-        set(CMAKE_BUILD_TYPE "Debug")
-    else()
-        set(CMAKE_BUILD_TYPE "Release")
-    endif()
+if(ENABLE_UNITTESTING)
+    include(CTest)
```
rianquinn: huh.... this might work. I would make sure that unit tests all run and pass based on both static and shared builds, but if this works, this is cleaner for sure.

bfsdk/cmake/CMakeGlobal_Project.txt
```diff
@@ -16,147 +16,33 @@
...
-# Prefix
-# ------------------------------------------------------------------------------
-
-message(STATUS "Install path: ${CMAKE_INSTALL_PREFIX}")
-
 # ------------------------------------------------------------------------------
 # OS
 # ------------------------------------------------------------------------------
 
+# TODO: Use CMAKE_SYSTEM_NAME to detect your host OS, and allow for build
```
rianquinn: That will not work on Cygwin as the toolchain is what defines this, and we need to know if we are on Windows / Cygwin, even if we get a toolchain that defines "Linux" which is our target for the cross compiled components.

scripts/cmake/config/default.cmake
```diff
@@ -0,0 +1,144 @@
...
+# ------------------------------------------------------------------------------
+# Developer Features
+# ------------------------------------------------------------------------------
+
+set(ENABLE_UNITTESTING OFF
+	CACHE BOOL
+	"Enable unit testing"
+	)
+
+set(ENABLE_ASAN OFF
```
rianquinn: All of these should be in the form:

```
set(ENABLE_ASAN 
    OFF
    CACHE BOOL "Enable clang AddressSanitizer"
)
```

rianquinn:
Also, I would love to move to ".cmake" instead of ".txt" as we make these changes as ".txt" is never picked up for syntax highlighting.

CMakeLists.txt
```diff
@@ -19,17 +19,16 @@
...
-
+include(${CMAKE_SOURCE_DIR}/scripts/cmake/global_vars.cmake)
+include(${CMAKE_SOURCE_DIR}/scripts/cmake/config/default.cmake)
 include(${CMAKE_SOURCE_DIR}/bfsdk/cmake/CMakeGlobal_Project.txt)
 
-# ------------------------------------------------------------------------------
-# Prefixes
-# ------------------------------------------------------------------------------
+if(ENABLE_DEPEND_UPDATES)
+    set(EP_UPDATE_DISCONNECTED OFF)
```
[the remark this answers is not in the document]
JaredWright: The built-in cmake directory variable EP_UPDATE_DISCONNECTED makes disconnected updates the default for all calls to ExternalProject_Add. This removes the need to declare     disconnected updates for each individual project. It does look like I actually set this variable wrong though, so I will need to change it slightly. If you think it makes more sense to have control over configuring each project for updates individually, then we can definitely put the individual UPDATE_DISCONNECTED things back.

CMakeLists.txt
```diff
@@ -39,24 +38,28 @@ set(BAREFLANK_INSTALL_TEST_PREFIX ${CMAKE_BINARY_DIR}/bfprefix_test)
 if(NOT WIN32)
 
     find_program(PYTHON_BIN python)
+    set(PYTHON_BIN ${PYTHON_BIN} CACHE INTERNAL "")
```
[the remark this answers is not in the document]
JaredWright: If you declare a cmake variable as "CACHE INTERNAL", it will remove it as a visible configuration option from cmake-gui and ccmake (the cmake curses gui).  I marked a lot of the variables that currently don't make sense as "configurable" CACHE INTERNAL. This particular variable will probably become configurable once I migrate PYTHON_BIN into an appropriate location in a toolchain or configuration file. That way you can configure a specific version of python if you so desire.

CMakeLists.txt
```diff
@@ -66,14 +69,8 @@ if(NOT WIN32)
 
 endif()
 
-find_program(NASM_BIN nasm)
```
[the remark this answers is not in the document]
JaredWright: There is another check for NASM in bfsdk/cmake/CMakeGlobal_Project.txt. Looks like I left it commented out on accident, I will put it back in for now. In a future PR, I would like to move the checks for NASM to an x86 specific toolchain file, as it would not make sense to require nasm if only building for ARM.

CMakeLists.txt
```diff
@@ -109,7 +105,6 @@ if(NOT WIN32)
         CONFIGURE_COMMAND   ""
         BUILD_COMMAND       ""
         INSTALL_COMMAND     ""
-        UPDATE_DISCONNECTED ${UPDATE_DISCONNECTED}
```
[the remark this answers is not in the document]
JaredWright: See comment above about the built-in cmake directory variable EP_UPDATE_DISCONNECTED

CMakeLists.txt
```diff
@@ -225,7 +217,7 @@ if(NOT WIN32)
     ExternalProject_Add_Step(
         binutils
         prefix_install
-        COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_BINARY_DIR}/external/binutils/prefix/bin/ ${BAREFLANK_INSTALL_PREFIX}/bin/
+        COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_BINARY_DIR}/external/binutils/prefix/bin/ ${BF_BUILD_INSTALL_DIR}/bin/
```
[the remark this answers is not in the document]
JaredWright: To me, the name BAREFLANK_INSTALL_PREFIX hints that this is the directory that will be used to install binaries/libraries/artifacts on the host system.  I propose the name     BF_BUILD_INSTALL_DIR for a couple reasons: I would like to use the prefix "BF_" to signify that this variable is shared across all sub-projects and is non-configurable (this is a fixed relative path in the build tree), and I think that the "BUILD_INSTALL_DIR" part of the variable name makes it more clear that this directory is used as the install directory FOR the build system. In a future PR, I would like to add a configurable variable (maybe a built-in cmake variable) that points to the directory for Bareflank tools to be installed on the user's system (i.e. when they run 'make install'). On a related note, I would also like to rename the variable BAREFLANK_INSTALL_TEST_PREFIX to a similar naming convention. That one seemed a little more difficult to migrate, so I left it for a future PR.

CMakeLists.txt
JaredWright: The variable BAREFLANK_HOST_STATIC_LIBS is always set to OFF under all build configurations, which makes it unnecessary in my opinion. You can decide if you need to build shared vs. static libs using only the variables BAREFLANK_HOST_SHARED_LIBS (now called BUILD_SHARED_LIBS) and the variable ENABLE_UNITTESTING. In future PRs, I would like to remove these global BUILD_SHARED and BUILD_STATIC variables in favor of providing configurable shared vs. static options for each individual project that produces a library,     with the option of building both if desired. You can do this with cmake by using a second call to add_library and configuring it for shared vs. static.

bfsdk/cmake/CMakeGlobal_Project.txt
JaredWright: Looks like I misread the docs on this one, good catch. I think that CMAKE_HOST_SYSTEM_NAME is the correct variable I was looking for.

scripts/cmake/config/default.cmake
JaredWright: Here are my reasons for wanting to use the format I suggested for configurable variables: First, I think that putting the "CACHE <TYPE>" part of the declaration on its own line makes it extremely easy to see the intended type of each configurable option. It is important to get these types right for all variables, because cmake-gui and ccmake use these types to determine how the variables are displayed. Second, by putting the description on its own line, it maximizes the space that the variable's author has to describe what the variable is for (assuming we wrap text at 80 chars?) For example, see TOOLCHAIN_PATH_GSL, which needs the whole line to contain an accurate description. Third, I think that putting the variable's name and default value on the same line looks more intuitive that you're assigning a variable that value. Maybe we can all fight about this on Gitter and come to a community consensus! :P

JaredWright:
It looks like there is some more work to do to make my changes compatible with Travis and Appveyor. Those will likely need to be updated along with the new build system as changes start coming in, looking into that now. Also, I apologize that that this PR contains so many changes at once. I'm hoping that it will be easier to make smaller PRs as more and more changes get integrated into the build system

CMakeLists.txt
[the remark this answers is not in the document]
rianquinn: Huh, I didn’t know that variable existed. That’s awesome. There should be no need to have an ENABLE_DEPENDS_UPDATES as you should be able to just state EP_UPDATE_DISCONNECTED, on the command line when configurng cmake.

CMakeLists.txt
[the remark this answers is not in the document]
rianquinn: Ah, thanks, yeah I don’t use cmake-gui, but I am glad that you are so that this is being tested.

CMakeLists.txt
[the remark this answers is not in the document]
rianquinn: Huh.... yeah your right. Originally the super-build didn’t include the project file so that nasm check was not there.

CMakeLists.txt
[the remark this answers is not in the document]
rianquinn: I am fine with BF_ so long as everything gets that prefix where appropriate. I like consistency :) . Your right that this is the install dir, but it is also a “prefix” in that, it’s the sysroot that we are constructing for the entire VMM. The hyperkernel will create other prefixs for guest software like our vmapps, and IncludeOS

CMakeLists.txt
rianquinn: Thats fine, so long as static and shared builds are being compiled as expected. I’m all for simplifying this, I just want to make sure all of the functionality is still there. In general, by default, the VMM and sysroot only compile static libraries and link to a single executable. If shared is stated, you would end up with shared libraries for the VMM and sysroot, but unit tests are still always compiled as static (unit tests always have to be static).

bfsdk/cmake/CMakeGlobal_Project.txt
rianquinn: Huh, if that works I would be very happy

scripts/cmake/config/default.cmake
rianquinn: That’s fine, but I’d like to compromise:

```
set(VAR <value>
    CACHE <type>
    “Description”
)
```

Right now, it appears that the indentation in the code (at least what GitHub is showing) is inconsistent, and in the rest of the code, we do not indent the “)”

rianquinn:
No issues with a large PR here as I would likely have had a similar issue. Changing the build system usually ends up with a lot of changes to get it right.

PR #511, opened 2017-10-11
astyle/clang_tidy: ignore deleted files

### LINKED ISSUE #509: bareflank_astyle_format.sh and bareflank_clang_tidy.sh try to process deleted files
Because these scripts use `git ls-files` to determine which files to process, they attempt to process files that have been deleted from the working tree but not from the index. `git ls-files` doesn't appear to have an option to exclude deleted files, but it does have one to emit _only_ deleted files, so the following somewhat long-winded replacement should work:

    git ls-files | sort | comm -23 - <( git ls-files -d | sort )

If we're not going to fix this we should at least document it; it took me way too long to figure out where they were getting the names of files I had removed...

cpavlina:
Fixes #509

### DIFF
--- a/bfsdk/scripts/bareflank_astyle_format.sh
+++ b/bfsdk/scripts/bareflank_astyle_format.sh
@@ -19,6 +19,10 @@
 
 OUTPUT=$PWD/.astyle_results.txt
 
+ls_not_deleted() {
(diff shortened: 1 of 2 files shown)


rianquinn:
LGTM

PR #512, opened 2017-10-12
Add PL011 serial driver, genericize portio interface

cpavlina:
This PR genericizes the portio interface to allow the serial driver to work under ARM (which uses MMIO), and adds a new driver for the [PrimeCell UART (PL011)](http://infocenter.arm.com/help/topic/com.arm.doc.ddi0183f/DDI0183.pdf) used in some ARM devices and in the current QEMU setup. The new driver is tested and working under QEMU (though still won't build for ARM without my "aarch64-hack" branch - check that out if you want to see it in action) and could compile and run on a hypothetical x64 system that used this device (so the tests run).

### DIFF
--- a/bfvmm/include/serial/serial_port_ns16550a.h
+++ b/bfvmm/include/serial/serial_port_ns16550a.h
@@ -16,14 +16,14 @@
 // License along with this library; if not, write to the Free Software
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
-#ifndef SERIAL_PORT_INTEL_X64_H
-#define SERIAL_PORT_INTEL_X64_H
+#ifndef SERIAL_PORT_NS16550A_H
+#define SERIAL_PORT_NS16550A_H
 
 #include <string>
 #include <memory>
 
 #include <bfconstants.h>
-#include <intrinsics/x86/common_x64.h>
+#include <intrinsics/common.h>
 
 // -----------------------------------------------------------------------------
 // Exports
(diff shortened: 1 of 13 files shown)


bfvmm/include/serial/serial_port_ns16550a.h
```diff
@@ -16,14 +16,19 @@
...
+#ifndef SERIAL_PORT_NS16550A_H
+#define SERIAL_PORT_NS16550A_H
 
 #include <string>
 #include <memory>
 
 #include <bfconstants.h>
-#include <intrinsics/x86/common_x64.h>
+
+#if defined(__x86_64__)
```
rianquinn: I'd love this logic to be in a header. Something like just including common.h, and it figures out which files you need. Otherwise, a lot of files are going to have this same logic, which will be annoying to fix later when we add AMD support

bfvmm/src/serial/src/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ include(${CMAKE_INSTALL_PREFIX}/cmake/CMakeGlobal_Includes.txt)
 
 list(APPEND SOURCES
     serial_port_ns16550a.cpp
+    serial_port_pl011.cpp
```
rianquinn: This should probably on be added to systems (like ARM) that need this hardware

bfvmm/src/serial/src/serial_port_pl011.cpp
```diff
@@ -0,0 +1,192 @@
...
+    return &serial;
+}
+
+void
+serial_port_pl011::set_baud_rate_divisor(uint32_t int_part, uint32_t frac_part) noexcept
+{
+    int_part &= 0xFFFF;
+    frac_part &= 0x3F;
+
+    // UARTIBRD and UARTFBRD aren't updated until UARTLCR_H is written, so save
```
rianquinn: Please move to a function level comment.

bfvmm/src/serial/tests/test_serial_port_pl011.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+TEST_CASE("serial: constructor_null_intrinsics")
+{
+    MockRepository mocks;
+    mock_serial(mocks);
+
+    CHECK_NOTHROW(std::make_unique<serial_port_pl011>());
+}
+
+
```
rianquinn: please remove extra spaces. If you could find a way to auto detect this, I'd love to see another CI test for this as it happens a lot (even from me)

bfvmm/include/serial/serial_port_ns16550a.h
cpavlina: This is what I mentioned in chat we'll have to discuss. `__x86_64__` isn't defined on MSVC. We need to decide how we're going to handle that (part of which is me reading the existing code to see if there's already something - there may be, I'll RFC when I'm more informed)

bfvmm/src/serial/src/CMakeLists.txt
cpavlina: I feel like it's adding unnecessary complexity to detect and branch on this. The PL011 code is written in an architecture-dependent manner - sure, only ARM systems have it, but it's not hurting anything sticking around. It's fairly small code too. If you really want to do this though, we should probably discuss how we'll go about switching on architecture in cmake - should probably include @JaredWright in that discussion too.

rianquinn: Point taken. Likely we will need something as we cannot compile all of the ARM logic on Intel and vice versa.

rianquinn: +1

bfvmm/src/serial/tests/test_serial_port_pl011.cpp
cpavlina: clang-format can do this with `MaxEmptyLinesToKeep`, we could always ditch astyle for clang-format since we use clang anyway. That'd require translating our astyle configuration though. If you prefer simple and cowboyish, the following awk script will print the number of extra blank lines in a file...

    awk '/^[ \t]*$/ {if (s) n++; s++} /[^ \t]/ {s = 0} END {print n||0}' < filename

rianquinn: We have used clang format in the past and it's not as good as astyle, so I would rather not change. Although the awk script is long, it would work fine for Travis. It would be nice if it would tell you which line numbers are the issue.

cpavlina: lol, I wasn't expecting you to be so okay with an ugly awk script. Here's a version that prints line numbers of extra lines - would be easy to pipe this through something else to format it nicely:

    awk '/^[ \t]*$/ {if (s) {n++; print NR;} s++} /[^ \t]/ {s = 0} END {exit (n != 0)}' < filename

It also exits nonzero if there are any extra lines so it can be used easily as a condition in a shell script.

rianquinn:
Ok. Please add this to make format. Make format should make the change but exit with non zero if a change has to be made like it does now.   
  

  
  

  
  
>

PR #514, opened 2017-10-13
Modify "make format" to check for extra blank lines

cpavlina:
This adds a script `bareflank_blank_lines.sh` which scans source files for consecutive blank lines and deletes extras, and modifies the `format` and `format-all` targets to call that script.


cpavlina:
False positive from codacy:

>Expressions don't expand in single quotes, use double quotes for that.

That's intentional, I need to include `$` in a string because it contains regular expressions...

PR #522, opened 2017-11-02
Move portio back into x64

cpavlina:
Rather than a generic portio interface for all architectures, with aarch64 implementing a MMIO-based fake portio, the one item that used portio generically (serial drivers) has been refactored to provide a generic implementation of its own IO which is conditionally compiled in whichever way is sensible for each architecture.


bfvmm/include/serial/serial_port_base.h
```diff
@@ -0,0 +1,166 @@
...
+///
+class EXPORT_SERIAL serial_port_base
+{
+public:
+#if defined(BF_X64)
+    using port_type = x64::portio::port_addr_type;
+    using value_type_8 = x64::portio::port_8bit_type;
+    using value_type_32 = x64::portio::port_32bit_type;
+#elif defined(BF_AARCH64)
+    using port_type = uintptr_t;
```
rianquinn: uintptr_t changes in size, what that intended? Also, does it make sense to make aliases for these in the intrinsics so that other modules are using the same type?

bfvmm/include/serial/serial_port_base.h
```diff
@@ -0,0 +1,166 @@
...
+    ///
+    /// @expects none
+    /// @ensures none
+    ///
+    /// @param str string to write
+    /// @param len length of the string to write
+    ///
+    virtual void write(const char *str, size_t len) noexcept;
+
+protected:
```
rianquinn: add newline

rianquinn:
Couple of questions but looks good

bfvmm/include/serial/serial_port_base.h
cpavlina: aarch64 uses MMIO for serial port IO, so the "port" type should be exactly the width of a pointer. As for making aliases in intrinsics - I don't think so, they're part of the abstracted IO specifically for use by the serial port drivers. Putting pieces of them in intrinsics seems like what I'm trying to remove in this commit.

rianquinn: +1

bfvmm/include/serial/serial_port_base.h
cpavlina: done

cpavlina:
Don't merge, forgot to add a few unit tests.

PR #526, opened 2017-11-03
Fix cast type in aarch64 serial MMIO methods

cpavlina:


rianquinn:
I'm assuming it didn't compile with static_cast? Also, why are you adding the qualifier?

cpavlina:
Indeed it does not compile with `static_cast`. In your second question are you referring to `volatile`? That's required for MMIO access; otherwise the compiler is free to assume consecutive accesses to the address will return the same value.

rianquinn:
I'm not sure that assumption stretches beyond the confines of a function, but I'm fine with the implementation. Once this passes I will merge.

cpavlina:
Depends whether the optimizer decides to inline the method - I've personally seen it do that. In any case it's "proper" to declare it that way whether or not the optimizer usually would mess with it.

rianquinn:
+1

Issue #527, opened 2017-11-03
[RFC] Passing platform-specific information to VMM from bfdriver


cpavlina:
# Description

On aarch64, I need to be able to pass information specific to the platform into the VMM from kernel space (in particular, the virtual address of the serial peripheral). This RFC proposes a way to handle this information.

# Proposed Changes

- In `bfsupport.h`, define `struct platform_info_t` with platform-dependent contents, and embed a member of this type into `struct crt_info_t` for passing into the runtime.
- In `bfplatform.h` and the various specific implementations, add `platform_populate_info(struct platform_info_t *)` which populates the struct with whatever information that platform requires.
- In `linux/platform.c`, `platform_populate_info()` on aarch64 should retrieve the serial base address and place it in the struct. All other implementations of this function (Linux non-aarch64, Windows, test) can be no-ops.
- In `common.c`, call `platform_populate_info()` inside `common_load_vmm()` before executing `call_vmm()` so that the VMM receives the information on `BF_REQUEST_INIT`.
- In `crt.cpp`, provide a function to return a pointer to this struct so that VMM code can access it.
- To the serial drivers, add a method `set_port()` to modify the port or base address at runtime.
- To `serial_port_base`, add a platform-specific constructor on aarch64 that fetches the `platform_info_t` and uses it to initialize the serial base address using `set_port()`. `set_port()` itself should remain a public method so that any VMM code that modifies the memory mapping may reassign the base address as it needs.

cpavlina:
I'm almost able to PR about half of this now; I want to ask a minor opinion question. Right now, we're using `DEFAULT_COM_PORT` on both x64 and aarch64 (for semantically the same thing, but technically different- one is the port IO base and one is a MMIO base address). To configure the mapping to be used by bfvmm prior to page table configuration, I need to `ioremap()` the peripheral in the kernel, which requires both a base address and a memory region size. I suggest using `DEFAULT_COM_PORT` only on x64, and on aarch64 using `DEFAULT_COM_ADDRESS`, `DEFAULT_COM_LENGTH`, and `DEFAULT_COM_DRIVER` --- the latter of which is the name of the driver to use (`pl011` or `ns16550a`). Any issues with that?

(Technically I could also drop `DEFAULT_COM_LENGTH` and just have the region length be a property the individual serial driver can return, but that's a bit more complicated to implement. Let me know though, I can do that if preferred.)

PR #528, opened 2017-11-03
Add method set_port() to serial drivers

cpavlina:
This allows changing the port at runtime, and will be used on aarch64 to support memory-mapped serial. Related: #527


cpavlina:
Will be a while before I address the appveyor build issues.

PR #531, opened 2017-11-21
Delegate Implementation

rianquinn:
This patch cleans up some issues with formatting, and adds support
for C++ delegates. This implementation doesn't need malloc/free
and is as fast as a regular vtable lookup. Before the release
of v1.2, a lot of the inheritance currently being used with be
converted to delegates to increase modularity.


rianquinn:
@JaredWright Please make sure this ends up in the new build system, it should if everything merges properly

JaredWright:
Merged without any issues, thanks for the heads up!

PR #536, opened 2017-11-30
Restructure bfvmm source and test files

cpavlina:
This commit implements a basic restructuring of `src` and `test` files within the bfvmm subtree, both to clean up build scripts for unit tests and to prepare for multi-architecture support. The following is implemented:

- `src/foo/src/*.cpp` (x64-related) -> `src/foo/arch/intel_x64/*.cpp`
- `src/foo/src/*.cpp` (non-specific) -> `src/foo/*.cpp`
- `src/foo/test/*.cpp` -> `test/foo/*.cpp`

No restructure of `include` files has been done yet - this is pending discussion on related code restructure (namespacing, etc). A similar split between architecture-specific and non-specific `test` code will be done in the near future similar to the organization of the `src` files. A similar restructuring of other subprojects is also planned for the near future.

I'm submitting this PR fairly early to facilitate the planned restructure of `bfvmm/*/intrinsics`.


JaredWright:
I think some of the bfvmm/test directories are going to need arch specific subdirectories. For example, you moved intel specific intrinics to bfvmm/src/intrinsics/arch/intel_x64, but intel specific intrinisics tests are under bfvmm/test/intrinsics. Is it possible/wise to make the test directory structure exactly mirror the src directory?

cpavlina:
Yes, that's the plan. I PR'd a working partial restructure now so the bfintrinsics restructure can commence without us messing with each other. I'll add that later.

connojd:
@cpavlina once @JaredWright's pending extended apis update gets merged can you restructure those as well?

cpavlina:
Sure, that shouldn't take long at all.

Issue #534, opened 2017-11-29
[BUG] build system - unnecessary re-compilation


JaredWright:
The build system re-compiles every file every time you build.

### Root Cause
This bug happens because some of the external dependency projects (hypervisor/scrips/cmake/depends/*.cmake) install their build artifacts into the build tree's install directory every time the project is built. The reason this happens is because the built-in cmake function ExternalProject_Add_Step is not correctly tracking dependencies of projects built using ExternalProject_Add.

### Steps to reproduce

1. Clone and build Bareflank in any valid configuration
2. Build bareflank again. ALL components will be re-compiled

### Temporary work-around

1. Clone and build Bareflank in any valid configuration
2. Run cmake again (i.e. ```cmake ..```)
3. Build again. Only source files that have been modified will be recompiled

JaredWright:
I may have found a fix for this... there is a built-in cmake command *copy_if_different* that you can use as a part of ExternalProject_Add_Step. You have to specify each file to be copied individually (doesn't work recursively), but it appears to fix the issue.

rianquinn:
Nice, git it a try and I can test once it's ready.

PR #539, opened 2017-12-01
[BUG] build system - unnecessary re-compilation

JaredWright:
Bug #534 

Added a macro and utility cmake script to allow installing files with
correct dependency tracking during build commands added with
ExternalProject_Add_Step

### DIFF
--- a/scripts/cmake/copy_files_if_different.cmake
+++ b/scripts/cmake/copy_files_if_different.cmake
@@ -1,0 +1,43 @@
+#
+# Bareflank Hypervisor
+# Copyright (C) 2015 Assured Information Security, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


connojd:
What should this patch be preventing? If it's the driver build duplication I think the issue is that there is an `execute_process` in the driver's CMakeLists.txt.  I ran this PR on my machine, and after commenting the execute_process out ithe driver wasn't rebuilt.

JaredWright:
This PR does not address the driver build duplication, maybe I should have named the bug slightly different. The behavior that this PR addresses is that all source files in the vmm, unit tests, bfm (basically anything that isn't the driver) re-compiles all source on every build. This was because external dependencies such as json, hippomocks, newlib, etc were being re-installed into the build "install" directory on every build, thus causing all components that depend on them to rebuild as well.

You are correct, the driver does rebuild itself every time because of the issue you described. I might suggest making seperate bug report for that one. Should I change the description of bug #539 to be more specific?

connojd:
Whatever is easier - If the driver fix is only that one line you could probably just throw it in with this one but you can make it separate if you want

JaredWright:
It will be a little more complicated that just commenting that line out, because that line is what causes the driver to build at all under the 'all' target. I think the real solution to driver-rebuilding is to remove the need for building the driver with an external build script (scripts/util/driver_build.sh). We should have cmake compile the driver project just like anything else, potentially using a driver-specific toolchain.

connojd:
👍

PR #551, opened 2017-12-11
[BUILD] Add USES_TERMINAL to targets

connojd:
This patch adds the USES_TERMINAL argument to the top-level
custom targets so that ninja no longer hangs on those targets


rianquinn:
@connojd CMake does not like your changes.

PR #554, opened 2017-12-11
Pack struct registers_intel_x64_t

cpavlina:
This is required for reliable access via assembly. We get lucky because the compiler is obviously going to align 64-bit members of a struct with nothing else in it to 64 bits on a 64 bit machine, but it's definitely not required to.


rianquinn:
LGTM

Issue #545, opened 2017-12-07
[RFC] Updated VMM Extension Integration


JaredWright:
## Overview

This RFC proposes a new mechanism for integrating the base Bareflank hypervisor repo with extensions (including the extended apis). I beleive that an ideal implementation would have the following characteristics:

* **Absolute minimal work required for the extension writer** - Developers should be able to start writing an extension by doing something as simple as including one file, or writing one line of code in their extension project.
* **Homogeneity** - The process for integrating the extended APIs and hyperkernel should be exactly the same as for custom developer extensions.
* **Out-of-tree** - Make it easy and desirable for Bareflank extensions to reside outside of the main hypervisor source tree. For example, it should not be necessary or encouraged to clone the extended APIs to hypervisor/eapis
* **Encourage inter-extension integration** - Extensions should be encouraged to integrate together, and depend on other extensions. For example, many people will probably want to use the extended APIs as a dependency of their own extension

Some other points to think about:
* Should we *require* extension writers to use cmake as the extension's build system, even if its just a minimal CMakeLists.txt file?
* Should the extension be built using the extension's build system, or should the we let the base Bareflank build system carry out the heavy lifting? Is it possible to allow for both options?

## Proposed API Changes

Here's an initial stab at how extension integration might work. If you'd rather read these changes in the form of source code, I have a mostly-working branch with these features implemented, and an example extension. To get my example branch working:

```
git clone -b extensions_overhaul https://github.com/JaredWright/hypervisor
mkdir hypervisor/build && cd hypervisor/build
cmake .. -DBFCONFIG=extensions_example.cmake
make -j<# cores>
```

### Extension Registration

Let extension writers declare their extensions to the base hypervisor build system using a build system configuration file (already part of the new build system), and a new *vmm_extension()* macro. The vmm_extension macro will support any arguments that cmake's built-in ExternalProject_Add macro supports. The hypervisor build system will do the heavy lifting, and "reach out" to the extension to clone it, build it, and do whatever else you declare in the configuration file. The hypervisor's build system will also copy the entire build configuration (cmake cache variables, etc) into the extension's scope. That way, the extension can access build configurations the exact same way the base hypervisor repo does.

An example build configuration might look like this:
```
# extensions_example.cmake

# First, you can override any default build configurations that you'd like:
set(BUILD_VMM_STATIC ON)
set(BUILD_VMM_SHARED OFF)
# etc...

# Next, you can add extensions to the build system using the vmm_extension
[...]

PR #560, opened 2017-12-13
New VMM extension integration mechanism

JaredWright:
1) New vmm_extension() macro for registering bareflank extensions to the
build system from build configuration file

2) New VMM extension-specific config variables and extension-specific macros

3) Added vmm_ex.cmake to be included by all Bareflank extensions

4) Updated build system documentation

[RFC] #545

### DIFF
--- a/scripts/cmake/config/bfconfig_template.cmake
+++ b/scripts/cmake/config/bfconfig_template.cmake
@@ -1,0 +1,58 @@
+#
+# Bareflank Hypervisor
+# Copyright (C) 2015 Assured Information Security, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+# ------------------------------------------------------------------------------
+# README
+# ------------------------------------------------------------------------------
+
+# Use this template and the examples below to configure the Bareflank build
+# system and add VMM extensions. Remove the '#' symbol to uncomment a line.
+#
+# For detailed instructions on configuring and building Bareflank, see
+# hypervisor/scripts/docs/build_instructions.md.
+
+# For detailed instructions on configuring and adding VMM extensions, see
+# hypervisor/scripts/docs/extension_instructions.md
+
+# To view all available build system configuration variables, use ccmake or
+# cmake-gui from your build directory:
+#      cd </path/to>/build; ccmake .
+
+# ------------------------------------------------------------------------------
+# Bareflank Build Configurations
+# ------------------------------------------------------------------------------
+
+# set(BUILD_TYPE Debug)
+# set(BUILD_VMM_STATIC ON)
+# set(ENABLE_DEVELOPER_MODE ON)
+
+# ------------------------------------------------------------------------------
+# Bareflank Extensions
+# ------------------------------------------------------------------------------
+
+# vmm_extension(
+#     extended_apis
+#     GIT_REPOSITORY https://github.com/bareflank/extended_apis.git
+#     GIT_TAG dev
+# )
+
+# vmm_extension(
+#     my_extension
+#     SOURCE_DIR ~/bareflank/my_extension
+#     DEPENDS extended_apis
+# )
--- a/scripts/docs/build_instructions.md
+++ b/scripts/docs/build_instructions.md
@@ -1,0 +1,140 @@
+# Configuring and Building the Bareflank Hypervisor
+
+## Dependencies
+
+Although Bareflank can be made to run on most systems, the following are the
+official supported platforms and their dependencies:
+
+* Arch Linux:
+```
+sudo pacman -S git linux-headers nasm clang cmake base-devel
+```
+
+* Ubuntu 17.04 (or Higher):
+```
+sudo apt-get install git build-essential linux-headers-$(uname -r) nasm clang cmake
+```
+
+Next, create a workspace to hold all Bareflank related source code, build
+artifacts and extensions...
+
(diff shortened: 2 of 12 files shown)


scripts/docs/build_instructions.md
```diff
@@ -0,0 +1,140 @@
+# Configuring and Building the Bareflank Hypervisor
+
+## Dependencies
+
+Although Bareflank can be made to run on most systems, the following are the
+official supported platforms and their dependencies:
+
+* Arch Linux:
+```
+sudo pacman -S git linux-headers nasm clang cmake
```
cpavlina: Also `base-devel`

scripts/cmake/config/bfconfig_template.cmake
```diff
@@ -0,0 +1,41 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+# ------------------------------------------------------------------------------
+# Bareflank Build Configurations
+# ------------------------------------------------------------------------------
+
+# set(BUILD_TYPE Debug)
```
rianquinn: We should provide documentation in the template so that all you have to do is read and modify the template.

scripts/cmake/config/default_ex.cmake
```diff
@@ -0,0 +1,129 @@
...
+# ------------------------------------------------------------------------------
+
+add_config(
+    CONFIG_NAME VMM_EX_TOP_DIR
+    CONFIG_TYPE PATH
+    DEFAULT_VAL ${CMAKE_CURRENT_SOURCE_DIR}
+    DESCRIPTION "Path to the top directory for this VMM extension"
+)
+
+add_config(
```
rianquinn: Are these optional? If not they should be, and our example should be the simplest possible config with options commented out with comments on how to use them if desired. Most of these seem like things we should be able to "infer" based on our "preferred" directory layout

scripts/cmake/config/default_ex.cmake
```diff
@@ -0,0 +1,129 @@
...
+    CONFIG_NAME VMM_EX_IS_UNITTEST_BUILD
+    CONFIG_TYPE INTERNAL
+    DEFAULT_VAL OFF
+    DESCRIPTION "The build system sets this variable to ON if this extension's "
+    "unit tests are currently being built, rather than it's source"
+    SKIP_VALIDATION
+)
+
+if(VMM_EX_IS_UNTTEST_BUILD)
```
rianquinn: Is there a way to simplify this? Everything else is pretty nice to look at, and seems easy to follow, this one part stands out as "yuk".

scripts/cmake/config/bfconfig_template.cmake
JaredWright: How much documentation should go in the template? I choose the commented out lines in the template as specific examples of 90% of what you would want to do with this file, namely:

* Setting a string type config
* Setting a boolean type config
* Adding an extension from GitHub
* Adding an extension from a local directory

What else should be added? Maybe a link or reference to the build instructions readme? One-sentence descriptions of what each section is for?

scripts/cmake/config/default_ex.cmake
JaredWright: The configs in this file are what get passed into the extension by the base build system, set to their default values. Extension writers can override these values in their extension-specific configuration files if desired. The configs here are exactly what you described: things we should be able to "infer" based on our "preferred" directory layout

scripts/cmake/config/default_ex.cmake
JaredWright: I agree that this is a bit "yuk", but I can't think of a simpler way around it. The real problem that this addresses is that there seems to be a pattern present in every VMM component (including the base components) that most of the source code needs to be re-built during a unit test build, but you still need to the ability to exclude a few things from a unit test build. With that in mind, I thought extension writters should have a variable available to check if the build system is currently building the extension normally, or if it was building the unit tests for the extension.

In the old build system, this same check was all over the place, but written as if(VMM_TOOLCHAIN_FILE AND ENABLE_UNITTESTING). I thought that was more yuk that having this variable available. 

Any alternative suggestions?

scripts/cmake/config/default_ex.cmake
JaredWright: After I read your comment again, I think you're viewing this file as the "template" config file for an extension. That was not the intent here, this is meant to be setup by the base build system and "passed into" the extension. Extensions then define their own config files to define more values, or override the ones here

JaredWright:
Another general comment I wanted to bring up about this PR:

I think that our build system is starting to cross over the line between normal-CMake and black-magic CMake. In other words, I just want to make sure that everyone is ok with the trade off of adding a non-trivial amount of complexity to our build system, with the payoff of ease-of-use for extension developers.

The only reason I bring this is that CMake has a steep learning curve for future maintenence, and we seem to be taking full advantage of all it has to offer.

scripts/cmake/config/bfconfig_template.cmake
rianquinn: I am a fan of redundancy, so in this case I would say adding the build instructions, and then a link to all of the options that are available in another doc would be fine so long as each option is documented somewhere.

scripts/cmake/config/default_ex.cmake
rianquinn: ok yeah, so I think the issue for me was the "_ex", as I read that as example. Please rename to "_extension" and you should be good here

scripts/cmake/config/default_ex.cmake
rianquinn: Your fine here as well since this is not something the user will be coding themselves.

scripts/cmake/config/bfconfig_template.cmake
JaredWright: How about adding something like this:

```
# For instructions on configuring and building Bareflank, see hypervisor/scripts/docs/build_instructions.md

# To view all available configurations, use ccmake or cmake-gui from your build directory:
#      cd </path/to>/build; ccmake .
```

I've been avoiding creating a file that documents all available build configurations because those things are going to change, get added, and get dropped really fluidly as we continue development. As the build README indicates, the best way to view all of the available configs is to use ccmake or cmake-gui. From those, you can see every available option, the possible values for each, and a description for each. If we go the README route for build configs, we'll have to remember to add/modify two things every time we touch a build config.

JaredWright:
Most recent commit adds the following:

1. Added more instructions to the template bfconfig.cmake file
2. Renamed all the "*_ex" cmake files to "*_extension"

Issue #557, opened 2017-12-13
[BUG] clang 5 vmm link error


cpavlina:
clang 5, which has now filtered down to our Travis builds, causes the following link error when building bfvmm:

    CMakeFiles/cxx_objects.dir/__/src/debug.cpp.o: In function `std::__1::__get_db()':
    /home/travis/build/Bareflank/hypervisor/build/depends/libcxx/src/src/debug.cpp:(.text+0x16a0): undefined reference to `__dso_handle'
    /home/travis/build/bareflank/hypervisor/build/install/vmm-x86_64-release/bin/x86_64-vmm-elf-ld: CMakeFiles/cxx_objects.dir/__/src/debug.cpp.o: relocation R_X86_64_PC32 against undefined hidden symbol `__dso_handle' can not be used when making a shared object
    /home/travis/build/bareflank/hypervisor/build/install/vmm-x86_64-release/bin/x86_64-vmm-elf-ld: final link failed: Bad value

cpavlina:
@connojd is currently looking into this; I'll take over from him if and when I finish what I'm doing now if he isn't done.

cpavlina:
I misread the errors and tried fixing this in a totally different place, as you can see by the PR I just opened and closed. Looks like the problem is actually in `libcxx`; I suspect the version we've cloned at [Bareflank/libcxx](https://github.com/Bareflank/libcxx) is no longer compatible with clang 5. We may need to upgrade it.

rianquinn:
Yes, but we don't want to do that right now as it's a pain. I'd prefer to wait on Clang 5 and merge in Clang 6 when it's ready. Moving to Clang 5 means updating libcxxabi, libcxx, newlib, and clang-tidy issues. 

For now, we should only support Clang 3.8, 3.9 and 4, which means in Travis CI, we will have to make sure that we are stating Clang 4.

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

Clang 5 is not supported. This fix prevents the build system from accidentally using it.

PR #561, opened 2017-12-14
TravisCI environment downgrade

JaredWright:
This commit forces the TravisCI environment back to its state before
the last environment upgrade happened. This does not fix compatilibility with
clang 5, but instead provides a temporary workaround for failing CI
builds.

[BUG] #557 - clang 5 vmm link error

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@
 #
 language: cpp
 dist: trusty
+group: deprecated-2017Q4
 
 #
 # Build Dependencies


rianquinn:
Although this fixes the issue, it doesn't fix the actual problem. In Travis CI, we are defining clang-4.0, but if you look at the build, it is using clang 5.0. We should be more explicit about what compile we want to use. If we state 4.0, it should use 4.0. 

I'm going to close this issue as I'd like to the underlying issue resolved instead.

Issue #553, opened 2017-12-11
[BUILD]: Double / Infinite Compile


rianquinn:
There are a couple of things that still double compile (or always compile):
- Newlib, Libcxx and Libcxxabi double compile. Once the double compile has been done, it doesn't compile again. This should only compile one. Either there is an issue with the headers being over-written when they shouldn't be, or the NASM issue has resurfaced with this new build system (if NASM is not defined properly in CMake, it will cause a double compile)
- The driver always compiles on every build.

rianquinn:
If I make a change to the VMCS, I get a recompile on modules that link to the VMCS. It should relink, but it should not recompile. 

<img width="1154" alt="screen shot 2017-12-11 at 10 33 39 am" src="https://user-images.githubusercontent.com/5932003/33844872-e6ffb01a-de5e-11e7-994c-122b41a5710a.png">

rianquinn:
I can also confirm that newlib compile every time you run make all until you run cmake a second time, which is clearly an issue. bfdriver compiles every time because make clean is being run.

JaredWright:
Here's my thoughts after looking into the numerous behaviors noted above:

1. Newlib doesn't really seem to double compile, its just that newlib doesn't use CMake, so it has its own mechanism for checking what does/doesn't need to be re-built (which takes forever...). I've observed the same behavior at random times with binutils (which also doesn't use cmake). I think the best solution to these issues is to implement the build system dependency caching feature we've been talking about. That way, the build system can cache the source code and/or build artifacts (libc, libc++, etc) and not have to worry about building those things at all (after you've done it once). I can open an RFC for that with more details.

2. The libcxxabi double-compiling appears to be resulting from newlib. Again, I think that dependency caching would be the best solution here.

3. The driver double-compile is fixed by removing the ```make clean``` line from the driver_build.sh script. I'd realllly like to get rid of those driver scripts all together, but I think we should push that off for another day...

4. I can't reproduce the screenshot bug for the VMCS changes recompiling modules that link against it. Mine relinks without recompiling.

rianquinn:
+1

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

PR #562, opened 2017-12-15
bfdriver double-compile bug

JaredWright:
Partial fix for [BUG] #553 Double/Infinite Compile. This commit removes
double compilation for the bfdriver component only.


rianquinn:
Merging regardless of Travis as this is a simple one

PR #571, opened 2018-01-08
Build system updates / Cleanup

rianquinn:
The following PR provides build system updates, while also provide the initial implementation some changes to the VMCS for using delegates instead of inheritance (lots more work to do here)


rianquinn:
bfvmm main is used to link everything from the prefix. It ensures that all of the libraries are installed and then links.

JaredWright:
Since there are too many changes to make comments inline, I thought I'd try making a list with permalinks to source on your branch. If there is a better way to do this, let me know:

1. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/macros.cmake#L317 This macro seems to be a duplicate of lines 309-314 in the add_dependency macro.

2. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L26 Any reason the add_config macro isn't in macros.cmake? Extensions should be able to use this macro as well.

3. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L388 Should we declare the valid options for this configuration so its easier to use from cmake-gui? (I think its something like ALLWAYS, LAZY, and NEVER). See https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L359 for an example

4. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L395 Same comment as 3 above

5. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L553 Does the build system need two VMM_TOOLCHAIN configurations now? I can't really tell what this is used for. Do some VMM components require regular ld and some ld.gold, or could we just have one VMM toolchain configuration variable, and let you configure the ld.gold as an alternate toolchain?

6. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L549 The default value here doesn't work for ARM toolchains. We could either change the name of the arm toolchains to match this pattern, or make the logic for calculating the name of the default toolchain a little smarter.

7. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/depends/json.cmake#L27 JSON is added twice when ENABLE_BUILD_TEST=ON

8. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/macros.cmake#L150 “Including dependency: ${name}” gets printed regardless of if the dependency will actually be included in the build. Might be a little confusing to someone who doesn’t know the details of the build system, and they might think that the dependency is being built instead of just "cmake included"

9. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/toolchain/native.cmake#L17 Wasn't 100% sure you intended this toolchain to be empty or not, so thought I'd point it out.

10. bfruntime and bfdummy don't have the src, include, test structure that all the other projects do

11. To me, bfdummy doesn’t belong at the top level, but this is just an opinion. Maybe in bfruntime/test? Or bfsdk?

12. bfm/main seems like it should be under bfm/src/main

13. Not all of the comments have been added back to the macros.cmake file.

14. What do you think about combining the “cache” dir and the “depends” dir into one? The tar.gz downloads could go to /depends/{dependency_name}/, and the source code could go to /depends/{dependency_name}/src/.

15. When the build system doesn't find a config.cmake in one of its preferenced locations, it used to copy the "template" config file (hypervisor/scripts/config/example_config.cmake) to build/config.cmake so that new users could have an easy-to-get-started place to both define configurations and add their extensions. Did that get removed on purpose? Could be added somewhere around this logic block: https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/config/default.cmake#L94

16. https://github.com/rianquinn/hypervisor/blob/625a30bf042d12888cef3134e3be900508e3a102/scripts/cmake/macros.cmake#L273 Wondering if it would be more clear to check PREFIX=“/path/to/a/prefix” instead of PREFIX == the string “VMM” inside all the macros

connojd:
https://github.com/rianquinn/hypervisor/blob/dev/scripts/cmake/config/default.cmake#L771 This should be userspace prefix path

rianquinn:
@JaredWright 

1: The add_dependency_cleanup is the “step” version of EPA with the prefix stuff encapsulated. 
2: Agreed
3: Agreed
4: Agreed
5: Yes. This extra tool chain is only needed for unit testing. Specifically, regular ld doesn’t create a relocation table for the PLT and GOT, it just creates a single table, while gold does the opposite. The best way so far to trigger both cases during unit testing has been to compile the unit tests using gold. In the future, we could probably spend some time and create fake relocation tables to support unit testing, but that will take some time. 
6: Agreed. Nice catch
7: Correct. One is for the VMM, and the other is for the TEST prefix. When unit testing is enable, both are needed. 
8: Agreed
9: I did, but I think I can do this a different way. 
10: Agreed
11: If it needs to be moved, it would actually have to be moved into the VMM as it is a fake VMM. The reason it is at the top level is because of the dependency tree. I’m sure there is a cleaner way to handle this, but for now, I think its fine. 
12: Agreed
13: Agreed
14: I don’t like that idea. The “make clean” targets simply remove the directories in question. Leaving the cache separate ensures that it’s never deleted accidentally. 
15: I did remove that on purpose. Not a huge fan of that. The config really needs to be provided during the original cmake process. If it’s not, and the user runs make, and then attempts to change a config, the new config variables are not distributed to the external projects. Instead, I think its a better idea to document how to create a config, and the encourage people to have the config ready to go during your initial configuration to prevent these types of issues. The same applies for ccmake BTW. It only works if you run ccmake before you run make. If you run make and then run ccmake, you run the risk of generating a corrupt build. 
16: it would as that would break the encapsulation part. The goal is to prevent the user from having to manually manage the prefixes. We also use the “name” part in the macros so doing so would make the macros needlessly more complicated.

rianquinn:
@connojd 

1: Disagree. This should be the VMM prefix as this binary can only be executed in the context of the VMM prefix. 
2: They always have been. I have no idea how to fix that. If someone knows how to I would love to know. 
3: There is a bug there

connojd:
@rianquinn re: 1: this is the output I get when I run ninja quick (I didn't see that the variable had VMM in it, but the bin directory still isn't being created under userspace-elf)

```bash
~/b/build $ ninja quick
[0/1] cd /home/cjd/bareflank/build && sudo /home/cj...k/build/prefixes/x86_64-userspace-elf/bin/bfm quick
[sudo] password for cjd: 
sudo: /home/cjd/bareflank/build/prefixes/x86_64-userspace-elf/bin/bfm: command not found
FAILED: CMakeFiles/quick 
cd /home/cjd/bareflank/build && sudo /home/cjd/bareflank/build/prefixes/x86_64-userspace-elf/bin/bfm quick
ninja: build stopped: subcommand failed.
```

JaredWright:
@rianquinn 

1. I think I was getting thrown off because the macro is named "add_dependency_cleanup". If you look at all the spots that macro is called (libcxx.cmake, binutils.cmake, etc), it seems like its being used to add additional steps to the dependency after it has been installed. Maybe we could rename it to "add_dependency_install_step", or something along those lines?

5. Maybe its just the name of the config thats non-intuitive to me. It sorta looks like you need two different toolchains for compiling the vmm, which really only needs one toolchain. Would VMM_TEST_TOOLCHAIN_PATH work for clarity?

15. I agree that the config should be provided before you run *make*, however I think that the process really should be "run cmake as many times as you want repeatedly until you get to config you like, and then run make". I think that the template config file encourages that mindset, and also gives you a copy/paste template for adding extensions to the build system right at your fingertips (instead of having to log into GitHub every time you forget the syntax of the add_vmm_extension macro). Finally, I liked that the template config file strongly encourages the "default convention" of putting a file named config.cmake right into your build directory. If you're already an expert with the build system (i.e. you already provide your own config files because you picked up on that convention), then the template file is invisible (it doesn't get copied if you already provided one).

rianquinn:
@connojd Yeah, I found that bug earlier. I will push a fix shortly

rianquinn:
@connojd Fixed. I addressed a couple of the issues addressed here. Once I get back, I will complete the rest.

rianquinn:
@JaredWright 

1. Agreed
2. Agreed
3. Disagree. Although I understand and agree with what your trying to do here, I feel it's added logic that doesn't provide enough value above and beyond some documentation on the readme to warrant the added complexity.

JaredWright:
I don't agree with the added logic and complexity argument, all of the logic is already there, you just add this to config/default.cmake, line 95:

```
execute_process(
    COMMAND ${CMAKE_COMMAND} -E copy_if_different
    ${SOURCE_CONFIG_DIR}/example_config.cmake
    ${BUILD_ROOT_DIR}/config.cmake
)
```
... and move that whole BFCONFIG import block down a few lines until its below the non-configurable variable definitions.

I would however gladly leave this alone if more people are opposed to having the template file placed there. @connojd @cpavlina: Do you have any opinions about having the build system put a template config file into the top-level of your build directory if one was not specified, and it doesn't already exist?

cpavlina:
I don't have any major opinions on that, to be honest. As discussed before, I'm _very_ much in favor of having a configuration template as I think it's vastly superior "documentation" to anything else we were going to provide, but as to whether it's copied into place automatically, I can't really say I care.

connojd:
I'm not opposed to having the template. The main thing I want is a TL;DR thing that shows you how to add an extension somewhere in the bareflank tree. One thing that I like about the checked-in config is that it makes the extension integration process the same for every extension. This allows us to get rid of the special if (ENABLE_EAPIS) checks.

rianquinn:
Agreed, so the template will stay for sure, and the documentation will point to the template, and how to use it (in addition to the template documenting itself). The only thing I'm opposed to is automatically copying the file to the build folder. The less we automate, the better. Besides the added code, I see myself having several config files with different names, but the build system basically will be forced to create this extra, unused config which just seems overkill for something we can document using a simple `cp <src> <dst>` command.

CMakeLists.txt
```diff
@@ -20,18 +20,30 @@ cmake_minimum_required(VERSION 3.6)
 project(hypervisor)
 
 # ------------------------------------------------------------------------------
-# Cmake global properties
+# Config
 # ------------------------------------------------------------------------------
 
 include(scripts/cmake/config/default.cmake)
+include_external_config()
```
JaredWright: I think this needs to be called before the default configs are included, or otherwise moved to an appropriate location high up in the default.cmake config file. Reason is that some of the "default values" rely on the values of other configs. In order for user-defined values to play nicely with the default values (and the add_config macro), they need to be defined first.

JaredWright: I think this needs to be called right *before* default.cmake is included to play nicely with the add_config macro, and all of the default values.

scripts/cmake/config/default.cmake
```diff
@@ -388,15 +326,17 @@ add_config(
...
+    OPTIONS ALWAYS LAZY NEVER
 )
 
 add_config(
     CONFIG_NAME CMAKE_BUILD_TYPE
     CONFIG_TYPE STRING
-    DEFAULT_VAL "Release"
+    DEFAULT_VAL Release
     DESCRIPTION "Defines the build type"
+    Release Debug
```
JaredWright: Missing "OPTIONS" keyword

scripts/cmake/config/default.cmake
```diff
@@ -783,21 +724,33 @@ add_config(
...
+        )
     else()
-        set(BFM_VMM "bfvmm_shared")
+        set(DEFAULT_BFM_VMM "bfvmm_shared"
+            CACHE INTERNAL
+            "Default BFM VMM"
+        )
     endif()
 else()
     if(BUILD_EXTENDED_APIS)
```
JaredWright: BUILD_EXTENDED_APIS isn't a defined configuration anymore (which is correct, it shouldn't be). I think we should change this to just point to bfvmm_shared or bfvmm_static as defaults, and let people overide the default to eapis in their configuration file.

JaredWright:
https://github.com/rianquinn/hypervisor/blob/8499fb459ec2197a0b42a77d058dc15793022f34/scripts/cmake/macros.cmake#L932 this variable is now is inconsistent with the BF_BUILD_COMMAND config in the targets file: https://github.com/rianquinn/hypervisor/blob/8499fb459ec2197a0b42a77d058dc15793022f34/scripts/cmake/targets.cmake#L53. It is also not available to the rest of the build system, because it is set in a function block without the PARENT_SCOPE option.

As a result, when you type ``` make info ```, it prints things like ``` driver_build - (description) ``` instead of ``` (make|ninja) driver_build  - (description) ```. Could we either put the build tool name back into that message, or otherwise at least print out something like ```"Available (make | ninja | msbuild) targets:```?

rianquinn:
Good catch. I still have to update that as code that was thee was completely broken for Windows. I wil add that to my list.

Issue #570, opened 2018-01-04
[BUG] build system: missing dependency for bfelf_loader headers


cpavlina:
There is a missing dependency relationship between the bfelf_loader headers in the source tree and their copies in the build tree, so if you modify the headers and try to rebuild, the modified files are not copied in and the build is done with stale files.

rianquinn:
My current mods fix this:
https://github.com/Bareflank/hypervisor/pull/571

I will close this ticket once we merge this code (could days away)

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

Issue #564, opened 2017-12-17
[BUILD]: Newlib disables build based on existence


rianquinn:
The following is used to gate the compilation of newlib:

```
if(NOT EXISTS "${BUILD_SYSROOT_VMM}/lib/libc.a" OR NOT EXISTS "${BUILD_SYSROOT_VMM}/lib/libc.so")
```

This needs to be removed as it will prevent the compilation of newlib if the upstream repo changes which means that people with an already established build will not get updates as needed.

JaredWright:
I would argue that this is the same as bug #553. Dependency caching will fix this as well (almost have an RFC and PR ready for that)

rianquinn:
I agree, but I don't want to forget about fixing this (and anything else like this) when caching is done

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

Issue #563, opened 2017-12-17
[BUILD]: Make distclean fails when unit testing is disabled


rianquinn:
When unit testing is disabled, make distclean fails because it attempts to clean code that doesn't exist. I have a fix for this in my current branch which I will post once it's ready.

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

Issue #535, opened 2017-11-29
[BUG] build system - windows support


JaredWright:
There a numerous issues preventing the new build system from compiling on Windows/Cygwin:

* bfdriver can't compile on windows. The .sln and .vcxproj files checked in to source control have invalid data (the software license) that prevents visual studio from using them
* A new toolchain for using msbuild and/or cl needs to be added for building bfm and bfdriver
* Compiler flags need to be added for msbuild/cl
* Appveyor needs to be updated to use the new build system. Add appveyor specific project configurations to scripts/cmake/config.

Randshot:
I have tried compiling and running the **dev** branch of [rianquinn's](https://github.com/rianquinn/hypervisor/tree/dev) fork.
The cmake configuration and the actual compilation work fine, but when I try to compile the driver with `make driver_build`, it throws the following errors:
```
hypervisor\bfsdk\include\bfdebug.h(990): fatal error C1083: Cannot open include file: 'wdm.h': No such file or directory
hypervisor\bfdriver\src\platform\windows\driver.h(22): fatal error C1083: Cannot open include file: 'ntddk.h': No such file or directory
platform.c(20): fatal error C1083: Cannot open include file: 'ntddk.h': No such file or directory
```

I have installed VS 2015, VS 2017 and the WDK.

rianquinn:
My changes are in. I still have some testing to do, but I think most of it is there. I updated the dev branch of hypervisor_example_cpuidcount so that you can see how it works. To use the config and extension.cmake, add -DCONFIG=<path> and -DEXTENSION=<path>

Randshot:
@rianquinn Nice. I will do some testing soon.

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/575

PR #575, opened 2018-01-21
Build System Bug Fixes

rianquinn:
This patch addresses several bugs with the build system:
- Lack of Windows support
- Clang Tidy was not working properly
- Making modifications to the source code would compile, but would not
re-link resulting in issues
- Removed excess file copying
- Make clean now works
- Removed unused or confusing folders in the build folder as part of the
build process. It is now much easier to traverse the build folder for
dependencies
- The targets didn't work with extensions. This has been fixed.

In addition, the following was also done under this patch
- Intrinsics was moved to the top level and out of the VMM
- Platform files are now in a folder called "platform" instead of
"arch"
- All subprojects have the same include/src/tests files structure
- All intel specific code is properly organized to match intended
namespacing
- Removed dead code
- A small part of the VMCS was converted to use the delegate pattern
- All dependencies are now downloaded at configure time instead of
compile time.
- The cache directory is now cached by Travis CI
- CMake output better matches old build system
- Added make rebuild and separate clean targets to remove various parts
of the build depending on needs.
- Added targets for Clean, Tidy, Rebuild and Format for each subproject
- Re-organized the cmake logic so that macros are not spread out
- New validation removes unneeded complexity
- Removed the need for the compiler wrapper, and in doing so, we now
provide a simpilar set of toolchain files
- Removed the need for Git repos. All external dependencies are
downloaded using a zip or tarball
- Libcxx and Libcxxabi are now in their own files. Much similar logic
- Unit test CMake files have been greatly simplified
- Each subproject is unaware of it's prefix and no long use VMM,
USERSPACE or TEST variables in their cmake files
- Fixed bugs with the flags
- Renamed the varbiables in the default.cmake config to be more
consistent and easier to follow in the rest of the code

https://github.com/Bareflank/hypervisor/issues/570
https://github.com/Bareflank/hypervisor/issues/564
https://github.com/Bareflank/hypervisor/issues/563
https://github.com/Bareflank/hypervisor/issues/557
https://github.com/Bareflank/hypervisor/issues/553
https://github.com/Bareflank/hypervisor/issues/535
https://github.com/Bareflank/hypervisor/issues/533
https://github.com/Bareflank/hypervisor/issues/529


Randshot:
@rianquinn So, I did some extension testing and it's working without problems.
I would like to mention though, that the **Ninja CMake Generator** doesn't seem to work on Windows (10).

Aside from that, I noticed that when running `make clean-all`, it seems to also run `cmake`.
```bash
$ make clean-all
Scanning dependencies of target clean-all
-- Config: /cygdrive/a/repos/bareflank_dev/example/config.cmake
-- Bareflank Manager VMM: example_vmm_shared
-- Including dependency: gsl
-- Including dependency: json
-- Including dependency: python
-- Including dependency: binutils
-- Including dependency: newlib
-- Including dependency: llvm
-- Including dependency: libcxxabi
-- Including dependency: libcxx
-- Extension: /cygdrive/a/repos/bareflank_dev/example/extension.cmake
--   ___                __ _           _
--  | _ ) __ _ _ _ ___ / _| |__ _ _ _ | |__
--  | _ \/ _` | '_/ -_)  _| / _` | ' \| / /
--  |___/\__,_|_| \___|_| |_\__,_|_||_|_\_\
--
--  Please give us a star on: https://github.com/Bareflank/hypervisor
--
--  Bareflank is ready to build, usage:
--     make
--
--  For more build options:
--     make info
--
-- Configuring done
-- Generating done
-- Build files have been written to: /cygdrive/a/repos/bareflank_dev/build
Scanning dependencies of target clean-prefixes
Scanning dependencies of target clean-depends
```

JaredWright:
@rianquinn I dislike the changes that slipped into this PR last minute concerning adding and building extensions from the build system. Is there any way to back those changes out to work how they did before this PR?

I'm concerned because you completely deleted a lot of well coordinated effort that many people contributed to and discussed via an RFC. I think that it is unfair to the community to undo all of that discussion and coordination without any warning.

rianquinn:
I can, however the code should have the same set of features, and it fixes a lot of issues with the previous code. I used the RFC as a reference for the work. The only difference is the name of the function (which I can always create the same function name and forward to add_subproject) and a Git repo is not supported as you should manage this yourself. We actually tried doing the whole git repo thing in one attempt with cmake and it was a total disaster as EPA was accidentally deleting your mods when it saw a change. Downloading your own repos and managing them yourself is by far the better approach. 

Also, I don't believe I "sneaked" anything in as I was pretty clear over gitter that I still had to cleanup extension work.

rianquinn:
Is there something in particular you are concerned with?

rianquinn:
I updated the comment to include which issues are corrected in this PR

rianquinn:
Also, I fixed the bug with clean-all.

JaredWright:
There are many more changes than what you just described, a couple examples:

* Extensions aren't added from the project config file (you can't add them there anymore). The only way to add extensions is from an extensions.cmake file at a hard coded location, or by passing via the command line an un-defined configuration variable named EXTENSIONS
* You removed all of the extension-specific configurations variables. I still think there are some configurations that will need to be extension specific
* All of the "conventions" and default behaviors we discussed around structuring an extension's source tree have been removed
* The suggested structure of the extension's tree is different (top level CMake file appears to call add_subproject, which calls into a nested src directory). This is inconsistent with the "src/include/test" structure of the other sub-projects. Maybe I'm just looking at the example cpuidcount extension wrong tho...
* From your example *cpuidcount* extension, it appears that extensions are now encouraged to over-ride the base hypervisor configurations (your example extension calls the bfm_default_vmm macro). This is going to lead to extensions competing to set the same set of build configurations. I don't think extensions should be able to override any configuration in the base build system

With regards to removing the ability to add extensions remotely via github: removing this feature drastically changes my plans around bareflank extensions that @cpavlina and I have been working on. If there is a good reason to remove this feature, I think we will need to coordinate it a little more.
 
Finally, I used the words "sneaking in" because you did mention "cleaning things up" via gitter, but that doesn't accurately describe what has happened, which is redoing 100% of that work in a PR that is already too large to feasibly review.

I'm trying not to sound too harsh with these comments, but I just want to make it clear that these changes are going to cause a lot of backtracking on the part of Chris and I.

connojd:
@rianquinn I agree with @JaredWright that the changes were "sneaked" in the sense that they were made under the general notion of "cleanup". "Cleanup" doesn't count as a clear discussion nor an RFC in my book. In my opinion, we either need to just scrap CONTRIBUTING.md for *everyone* or enforce CONTRIBUTING.md for *everyone*.

One possible way this could be done is through server-side git hooks. I think we should write hooks that will reject any PR that have over X changes.

Randshot:
@JaredWright 
I have experimented with the current (PR) extension system a bit and extensions don't necessarily change the default vmm.
I think it is a bit confusing, since the `config.cmake` is in the extension repo and not in the build directory.

I myself have structured my test extension like this.
**config.cmake (in the tlbsplit extension repo):**
```cmake
# Set the VMM
set_bfm_vmm(tlbsplit)

# Extensions
set(EXTENSIONS
    ${CMAKE_CURRENT_LIST_DIR}/../extended_apis/extension.cmake
    ${CMAKE_CURRENT_LIST_DIR}/extension.cmake
)
```

**${CMAKE_CURRENT_LIST_DIR}/extension.cmake:**
```cmake
add_subproject(
    tlbsplit vmm
    DEPENDS bfvmm bfintrinsics eapis
    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/src
)

add_subproject(
    monitor userspace
    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/monitor
)
```

**${CMAKE_CURRENT_LIST_DIR}/../extended_apis/extension.cmake:**
```cmake
add_subproject(
    eapis vmm
    DEPENDS bfvmm bfintrinsics
    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/src
)
```

I am not saying that it is necessarily better this way, I just wanted to clarify how I understand the current extension system.

rianquinn:
> Extensions aren't added from the project config file (you can't add them there anymore). The only way to add extensions is from an extensions.cmake file at a hard coded location, or by passing via the command line an un-defined configuration variable named EXTENSIONS

That was broken in the previous code as config changes were not happening in all cases. Since the previous code added the extensions after the projects main subproject, some of the variables were being missed. To fix that, you would have to "queue" up all of the external project add functions which is logic that is not needed if you simply provide two different cmake files (one for config, and one for your extensions), which also allows you to setup more than one config for the same extension during testing. 

Also, the locations are not hardcoded. You can provide them in your config file just like you can on the command line, and they are brought in using a list so you can have more than one. All of the same functionality is there. The only hardcoded location is the top level file that was there in the previous code for configs. 

> You removed all of the extension-specific configurations variables. I still think there are some configurations that will need to be extension specific

Like? This was not documented in the RFC

> All of the "conventions" and default behaviors we discussed around structuring an extension's source tree have been removed

Not true. It's the same. In the RFC you have a configuration file, and then a project file. Each extension can have one or more of each. The example I provided follows that same convention. 

> The suggested structure of the extension's tree is different (top level CMake file appears to call add_subproject, which calls into a nested src directory). This is inconsistent with the "src/include/test" structure of the other sub-projects. Maybe I'm just looking at the example cpuidcount extension wrong tho...

You are and you aren't. https://github.com/Bareflank/hypervisor/blob/1e27775318b1b64f515de65abe1b554bbf7b53e4/bfvmm/CMakeLists.txt

The project file is still there, it's simply not in the "src" directory. That is something I can change on the main repo. It doesn't really matter where the CMakeLists.txt is so long as one exists and you can point to it. In my latest change, I added this so that at least the API is identical:

```
function(vmm_extension NAME)
    list(APPEND ARGN
        DEPENDS bfvmm bfintrinsics
    )

    add_subproject(
        ${NAME} vmm
        ${ARGN}
    )
endfunction(vmm_extension)
```

This is helpful, but you can see why it's a bit redundant. 

> From your example cpuidcount extension, it appears that extensions are now encouraged to over-ride the base hypervisor configurations (your example extension calls the bfm_default_vmm macro). This is going to lead to extensions competing to set the same set of build configurations. I don't think extensions should be able to override any configuration in the base build system

They can, and need to be able to. Calling that macro is the same as the following code that you put in your RFC:

```
set(BFM_DEFAULT_VMM jareds_example_vmm)
```

All that macro is doing is adding _shared and _static depending on the build properties. The fact that the extensions and config are separated BTW allows you to have more then one config for the same set of extensions, which will be used a lot during testing for Travis CI as we can bring in different configurations without having to re-write how the extensions are setup. 

Also, this comment was also not articulated in the RFC. 

> github: removing this feature drastically changes my plans around bareflank extensions that @cpavlina and I have been working on. 

I will add this functionality back to the download_dependency function so that you can download a Git repo, but I HIGHLY discourage it's use.

@connojd Like I said before, I do not believe an RFC is needed for this PR as my goal with this is to maintain the same functionality that the previous RFCs proposed. The RFC that is missing is the delegate RFC, which I will write-up today. If there is something that is wrong with my implementation of the RFCs please let me know and I will correct it. 

Also, the word "cleanup" has been removed so that it is explicit what was changed. The reason for the new PR was specific to make it clear what was changed.

JaredWright:
I think I am failing to convey the single message that I'm trying to get through: This PR has reached a level of complexity that we can neither communicate nor coordinate everyone's needs in regards the feature set being added. My annoyances are coming from the fact that it seems that there is no upper bound on the amount of changes that are continuing to be added.

With that in mind, I can see that any technical feedback I offer will be neither constructive nor useful at this point, because I can't fully comprehend the entirety of this change set. Perhaps it will be best to put my needs aside for now, and offer my feedback in the form of numerous pull requests in the future.

It appears that the path of least resistance going forward is to just merge this thing and address issues as they arise. No point in me trying to respond to every single comment here.

rianquinn:
I added the following:
```
download_dependency(
    mytest
    GIT_REPOSITORY https://github.com/Bareflank/GSL.git
    GIT_TAG master
)
```

Downloads are placed in the cache along with the rest of the dependencies. If you want to script pulling down a subproject (also a bad idea), you can simply add the following to your config

```
execute_process(COMMAND git clone ${ARG_GIT_REPOSITORY} ${ARG_GIT_TAG} ${SRC})
```

Which will let you place the repo anywhere you want. Anything else? I'm going to move the CMakeLists.txt so that they are consistent with the example extension. Other than that, IMO, this PR implements the existing RFCs as intended while fixing a lot of issues, and even removing almost 11K lines of code. I wish it wasn't such a large PR, but what is done is done. 

The only thing that is not working are the unit tests. They run fine, but some are missing, and codecov is not getting the results correctly which I will be working on today

rianquinn:
Ok, that should be all of the requested fixes

JaredWright:
No further comments or requests from me, 👍  to merge

rianquinn:
I updated the example: 
https://github.com/Bareflank/hypervisor_example_cpuidcount/blob/dev/extension.cmake

Randshot:
@rianquinn The `add_subproject` macro still exists, right?
I am asking because of userspace applications.

rianquinn:
Yes, the vmm_extension macro is just a wrapper. I should have the code merged soon, I'm just trying to fix a bug I introduced this morning

rianquinn:
@Randshot is working on porting the EAPIs so my next tasks is the delegate RFC and getting the unit tests working again. I'd also like to start and RFC on moving to master and what people would like to see before we click merge.

Issue #576, opened 2018-01-22
[BUILD]: Unit tests not working properly


rianquinn:
The following issues still exist with the unit tests:
- Not all of the code is currently being tested. This is the result of moving some code around, as well as some code still needing to be ported to use the new intrinsics. 
- Some tests crash on WIN32 builds. Not sure why yet. 
- Codecov is not getting any reports which is why it says everything is fine. Likely, its just not configured properly. My suspicion is that it has not been working for a while.

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/582

PR #578, opened 2018-01-22
Fix crash with Windows

rianquinn:
The latest versions of Windows do not like having aligned memory freed
using "free", resulting in a segfault. Alighed memory is no longer
needed so this can be removed.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/576


rianquinn:
@connojd I think this one is G2G as well. Travis has been terrible over the past couple of days. At least 2-3 tests fail to even start, which is why it "failed". I just restarted it.

Issue #573, opened 2018-01-11
[BUG] VMM freezes on stop with Linux 4.14


connojd:
There is a bug in the promote.asm code that causes the VMM to freeze. The bug behaves similarly to the one #486 fixed.

rianquinn:
Fixed

PR #584, opened 2018-01-28
x64/vmcs: Temporarily clear CR3[11:0] on promote

connojd:
As part of the PTI patches, Linux now uses PCIDs to reduce the
cost of context switches. The current PCID is given by CR3[11:0],
and if this is not zero when CR4.PCIDE is set on a write to CR4,
a #GP is raised.

The promotion bug described in #573 was due to Bareflank
restoring the guest CR3 (in which CR3[11:0] was nonzero), and
then restoring the guest CR4, which was setting CR4.PCIDE. The
resulting #GP was raised with nothing to handle it, leading to a
triple fault.

Now Bareflank writes a copy of the guest CR3 with bits 11-0
clear, then writes the guest's CR4. Once the CR4 is
written, the guest's original CR3 is restored.

### DIFF
--- a/bfvmm/src/hve/arch/intel_x64/vmcs/vmcs_promote.asm
+++ b/bfvmm/src/hve/arch/intel_x64/vmcs/vmcs_promote.asm
@@ -174,13 +174,33 @@ vmcs_promote:
     vmread rdi, rsi
     call _write_cr0 wrt ..plt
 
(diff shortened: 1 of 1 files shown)


rianquinn:
LGTM +1

connojd:
Fixes #573

Issue #587, opened 2018-01-30
[BUG]: ccmake shows variables that cannot be changed


rianquinn:
Certain variables cannot be changed by ccmake, but they show up in ccmake regardless.

rianquinn:
Fixed: https://github.com/Bareflank/hypervisor/pull/588

PR #588, opened 2018-01-30
Remove CMake variables that cannot be changed in CCMake

rianquinn:
This patch makes sure that the ccmake tool only shows variables that in
theory should be able to be changed prior to running make.

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/587

### DIFF
--- a/scripts/cmake/config/default.cmake
+++ b/scripts/cmake/config/default.cmake
@@ -306,11 +306,22 @@ add_config(
 # Config
 # ------------------------------------------------------------------------------
 
+set(CMAKE_BUILD_TYPE "Release"
+    CACHE INTERNAL
+    "Defines the build type"
+)
+
+set(CMAKE_VERBOSE_MAKEFILE OFF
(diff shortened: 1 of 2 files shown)


JaredWright:
Only a couple comments:

1) The commit message uses the word cleanup

2) Are you sure we want to mark the toolchain configurations as advanced? This might make cross-compiling for ARM tricky

3) ccmake does indeed appear to be working on my environment, and the CMakeCache is being updated correctly as well. Was the root cause from using the add_config macro to define CMAKE_* variables?

👍 for getting ccmake working again!

rianquinn:
@JaredWright 

1: Fixed
2: Advanced only effects ccmake. If your going to change a toolchain, it needs to be done on the command line. That is a cmake limitation
3. Awesome. That fix I added was only there to fix an issue with VERBOSE not having a default value. To be honest, I'm not sure what changed as it was working fine when I tried it (outside of the variables that you cannot change)

Your welcome 😄

PR #604, opened 2018-02-08
The Great Merge of 2018

JaredWright:
This PR requests hypervisor/dev to be merged into hypervisor/master. Here is the strategy that I used to get this merge together:

```
git checkout master
git merge -X theirs dev
git rm (all files marked “Deleted by them”)
git merge —-continue
```

@rianquinn I'll let you pull the trigger on this one if/when you think the time is right. If we do this merge, we will probably want to tag it as well


connojd:
+1 for the tagging. I would also like to start including the shortlog every time we tag a release from here on out. That would give a nice concise overview of who committed what

rianquinn:
Yeah, tagging will be needed. The merge cannot happen until the EAPIs and examples have all been ported. Once Connor and I are complete with the delegate work, we should be able to hammer out the rest pretty fast.

PR #605, opened 2018-02-09
Add changes needed by eapis port

connojd:


connojd:
@rianquinn this will need to be merged before the extended apis. Once it is I will point the travis and appveyor of the eapis to use dev (or master if #604 is merged before this).

rianquinn:
@connojd Is this still needed?

connojd:
@rianquinn yes but it is out of date. Ill close and then reopen with updates

PR #617, opened 2018-02-16
Add IntelliSense Support

rianquinn:
This patch adds support for IntelliSense for VSCode. Specifically, this
provide support for code completion and error checking.

### DIFF
--- a/scripts/.vscode/c_cpp_properties.json
+++ b/scripts/.vscode/c_cpp_properties.json
@@ -1,0 +1,98 @@
+{
+    "configurations": [
+        {
+            "name": "Mac",
+            "includePath": [
+                "/usr/local/include",
+                "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1",
+                "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.0.0/include",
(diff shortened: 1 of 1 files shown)


JaredWright:
Can this be placed in the scripts directory instead of the top level?

rianquinn:
No, it has to be at the top level, a problem consistent with most editors.

rianquinn:
This is a feature @brendank310 and I have been trying to add for a long time, but nothing has worked until now. VSCode is pretty awesome.

JaredWright:
The problem that I see is that it is very specific to your development environment. To me, this doesn't belong checked in to the top level.

rianquinn:
It's not the first code completion file that we have checked into the top level. If we end up with more than one in the future that actually works, I am willing to consider a different approach.

JaredWright:
I feel like if I made a PR to check in a .vimrc to the top level of Bareflank/hypervisor with code completion support, there would be push back from people who don't use vim...

rianquinn:
Not from me

JaredWright:
I'm still not a fan of this PR. If you merge this, I'll likely suggest adding a makefile target that will set up VSCode features if you so desire. I would hate to see .eclipse and .xcode and .vim and .<every_other_ide> polluting the top level of the hypervisor repo.

rianquinn:
I moved it to scripts. I personally don't think this is an issue, but this is not something worth arguing over.

PR #620, opened 2018-02-18
Add support for unordered_map

rianquinn:
This patch adds support for unordered_map. Specifically it adds support
for libm which is needed by unordered_map as it abuses the ceil
function for integer rounding in it's hash function.


rianquinn:
Will have to add compiler-rt for this to work. Going to close for now, and I will reopen when I have this added.

PR #621, opened 2018-02-18
Delegate and KVM bug fixes

connojd:
* `delegate::is_valid` now compares `m_obj != nullptr` in addition to `m_stub != nullptr`. This allows         users to use 

  ```cpp
  auto d = delegate<FOO>::create<nullptr>();
  ```
  to initialize invalid delegates. Without the `m_obj check`, the delegate is created and `is_valid()` but   still has a nullptr `m_obj`

* Created a `bfvmm::intel_x64::kvm_guest` namespace that contains alternative rdmsr and wrmsr handlers that is specific to running as an L1 KVM guest. The choice between the original and alternative is made in the exit handler constructor. I figured this was better than checking for the KVM case on every exit in the original handlers themselves.

If any has any ideas or feedback on how to handle these situations that would be great

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ bfdriver/src/.common.o.cmd
 bfdriver/src/platform/linux/.bareflank.ko.cmd
 bfdriver/src/platform/linux/.bareflank.mod.o.cmd
 bfdriver/src/platform/linux/.bareflank.o.cmd
+bfdriver/src/platform/linux/.cache.mk
 bfdriver/src/platform/linux/.entry.o.cmd
 bfdriver/src/platform/linux/.platform.o.cmd
 bfdriver/src/platform/linux/.tmp_versions/
--- a/bfsdk/include/bfdelegate.h
+++ b/bfsdk/include/bfdelegate.h
@@ -78,7 +78,7 @@ public:
     /// @return true if valid, false otherwise
     ///
     constexpr bool is_valid() const noexcept
-    { return m_stub != nullptr; }
+    { return m_stub != nullptr && m_obj != nullptr; }
 
     /// Is Valid
     ///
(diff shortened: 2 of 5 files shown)


rianquinn:
I like the additions to the delegate implementation, but I am not a fan of the KVM specific handler. In general, not all systems will have the global perf MSR, and so we should have code in the generic handler that can handle when this is not present. I would prefer to see a patch that fixes the issue outright, and that is still performant.

connojd:
The easiest change is to check if the field exists and the msr allows the control to be set to 1. Previously you said you didn't want to check if the field exists in the msr handlers because its too slow, which I agree. What implementation do you want to see?

rianquinn:
We should check if the field exists during the construction of the exit handler and set a global variable that can be checked when this field is accessed.

rianquinn:
Might not be a bad idea to create a set of booleans for various different MSRs, CPUIDs, etc... that tell us if a specific feature exists or not. Something like a bfvmm::intel_x64::features class that can be created, and passed around in the exit handlers, so that they can check this for which features exist, and then we can add to that over time as needed

Thoughts?

connojd:
I assume a class like that would be essentially read-only after init? I main advantage I see of the additional class over the current exists/allowed logic: Enumerate available features once during initialization. This will save you from vmreading/vmwriting every time you need to check a feature. I'm not sure of the performance difference between those instructions and a memory read, but that approach will certainly be faster in the nested case.

rianquinn:
Yeah, They should be fast enough, and I like the read-only approach. Basically, you should have a bunch of "inline constexpr auto foo() const" functions that do nothing but return a bool that is initialized during construction. That should reduce it down to a single memory read, that should be cached if the read occurs a lot (which it won't for a lot of these)

PR #623, opened 2018-02-18
Change order of execution of delegates from FILO to FIFO

boddumanohar:
Signed-off-by: "Manohar Reddy" <"201451024@iiitvadodara.ac.in">

### DIFF
--- a/bfvmm/include/vcpu/vcpu.h
+++ b/bfvmm/include/vcpu/vcpu.h
@@ -260,7 +260,7 @@ public:
     /// Add Run Delegate
     ///
     /// Adds a run delegate to the VCPU. The delegates are added to a queue and
-    /// executed in FILO order. All delegates are executed unless an exception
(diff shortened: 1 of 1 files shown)


connojd:
Is there a reason not change the exit handlers as well?

boddumanohar:
https://github.com/Bareflank/hypervisor/blob/dev/bfvmm/include/hve/arch/intel_x64/exit_handler/exit_handler.h#L119

I guess exit handler already uses FIFO

rianquinn:
Yeah, the exit handlers were already FIFO. I think I changed the exit handlers, and forgot to change this one.

boddumanohar:
Looks like I broke the build. It built in my local machine. I checked it before pushing it.

connojd:
yeah windows is very touchy

rianquinn:
It appears that Cygwin is broken at the moment. I'll merge as this should get cleared up by Cygwin in the near future (usually does).

PR #624, opened 2018-02-18
README.md: Update Arch instructions

connojd:
- Change 'sync-deps' to 'syncdeps'
- Add clang-tidy-4.0 symlink


rianquinn:
Why are we manually installing yaourt instead of installing this via pacman?

connojd:
because yaourt isn't an official arch package

PR #626, opened 2018-02-19
x64/hve: Fix KVM and promote bugs

connojd:


bfvmm/src/hve/arch/intel_x64/exit_handler/exit_handler.cpp
```diff
@@ -29,6 +29,8 @@
 #include <memory_manager/memory_manager.h>
 #include <memory_manager/arch/x64/root_page_table.h>
 
+bool g_guest_perf_glbl_ctrl_field_exists;
```
rianquinn: For now, lets get rid of the global variable, and incur the lookup cost as this will not execute enough to matter right now. Once we have a "feature" class we can do this a little more cleanly.

bfvmm/src/hve/arch/intel_x64/exit_handler/exit_handler.cpp
```diff
@@ -141,7 +143,11 @@ handle_rdmsr(gsl::not_null<bfvmm::intel_x64::vmcs *> vmcs)
             break;
 
         case ::intel_x64::msrs::ia32_perf_global_ctrl::addr:
-            val = ::intel_x64::vmcs::guest_ia32_perf_global_ctrl::get();
+            if (g_guest_perf_glbl_ctrl_field_exists) {
```
rianquinn: Instead of using a global variable, just go the lookup here

bfvmm/src/hve/arch/intel_x64/exit_handler/exit_handler.cpp
```diff
@@ -317,6 +327,8 @@ exit_handler::exit_handler(
         }
     }
 
+    g_guest_perf_glbl_ctrl_field_exists = ::intel_x64::vmcs::guest_ia32_perf_global_ctrl::exists();
```
rianquinn: not needed anymore

rianquinn:
Please resubmit for "master"

PR #632, opened 2018-02-26
Add platform_info_t, support runtime serial port address

### LINKED ISSUE #527: [RFC] Passing platform-specific information to VMM from bfdriver

cpavlina:
platform_info_t is used to pass platform-specific information from the kernel into the vmm. On aarch64, this is used to map the serial peripheral and pass the resulting virtual address.

Closes #527

It is possible we may want to restructure the small section of code in `platform.c`, which is currently implemented using `#ifdef`; we could of course pull this out into `arch/intel_x64/platform.c` and `arch/aarch64/platform.c` or something - let me know if you want me to make changes.

### DIFF
--- a/bfdriver/src/platform/linux/platform.c
+++ b/bfdriver/src/platform/linux/platform.c
@@ -17,6 +17,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
+#include <bfarch.h>
 #include <bftypes.h>
 #include <bfdebug.h>
 #include <bfplatform.h>
@@ -31,6 +32,10 @@
 #include <linux/cpumask.h>
 #include <linux/sched.h>
 #include <linux/kallsyms.h>
+
+#if defined(BF_AARCH64)
+#   include <asm/io.h>
+#endif
 
 #include <asm/tlbflush.h>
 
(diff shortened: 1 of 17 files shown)


bfdriver/src/platform/linux/platform.c
```diff
@@ -195,3 +200,42 @@ platform_restore_preemption(void)
...
     put_cpu();
 }
+
+#if defined(BF_AARCH64)
+
+int64_t
+platform_populate_info(struct platform_info_t *info)
+{
+    info->serial_address = (uintptr_t) ioremap(DEFAULT_COM_PORT, DEFAULT_COM_LENGTH);
+
```
rianquinn: no need for whitepsace here

cpavlina:
Decreased coverage? Thought I got everything! ohno.png

bfdriver/tests/test_common_load.cpp
```diff
@@ -174,4 +174,22 @@ TEST_CASE("common_load_vmm: add tss mdl fails")
...
+    binaries_info info{&g_file, g_filenames_success, false};
+
+    for (const auto &binary : info.binaries()) {
+        REQUIRE(common_add_module(binary.file, binary.file_size) == BF_SUCCESS);
+    }
+
+    CHECK(common_load_vmm() != BF_SUCCESS);
+    CHECK(common_fini() == BF_SUCCESS);
+
+    platform_info_should_fail = 0;
```
rianquinn: use a gsl::finally instead as it is safer.

cpavlina: oh my, that's kind of embarrassing to be honest. Yes, I should know better :P

cpavlina:
Not sure why Codacy is unhappy...?

PR #636, opened 2018-03-02
Master preserved

connojd:
This patch reverts 6fda0bc1..4b2c963b, and then re-merges dev with a merge commit in order to preserve history


connojd:
@rianquinn @JaredWright I belive this is good to go.

PR #642, opened 2018-03-08
Move Intel-specific vcpu code to hve

connojd:
Downstream projects like the extended APIs need a stable interface. To
facilitate this, the existing interface will be changed to pass a single
vcpu pointer to client code. Any state or functionality needed by
extensions will be reachable from the vcpu pointer.

This design requires forward declarations of the vcpu class, resulting
in circular references. While this works on Linux, Windows does not
allow inter-dll circular references.  The workaround is to place the
API's vcpu code into hve.

Move the Intel-specific vcpu code to hve to remain consistent
with the extended APIs.

### DIFF
--- a/bfvmm/include/hve/arch/intel_x64/vcpu/vcpu.h
+++ b/bfvmm/include/hve/arch/intel_x64/vcpu/vcpu.h
@@ -19,11 +19,31 @@
 #ifndef VCPU_INTEL_X64_H
 #define VCPU_INTEL_X64_H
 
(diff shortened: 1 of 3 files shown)


rianquinn:
@connojd Remove the .cpp and put it back to a .h only. That will fix your codecov issue, and it will remove the need to link in the vcpu in hve, which I want to avoid.

connojd:
@rianquinn should be ready once CI passes

PR #645, opened 2018-03-08
x64/hve: Fix KVM and promote bugs

connojd:
exit_handler.cpp:
KVM does not allow the "load IA32_PERF_GLOBAL_CTRL" to be set to 1 in
both the vm-exit and vm-entry controls. This prevents the vmcs field
guest_ia32_perf_global_ctrl from existing, so reads and writes
to it (e.g. handle_rdmsr and handlr_wrmsr in exit_handler.cpp) cause an
exception to be raised by Bareflank.

- Modify the bfvmm's exit_handler private functions to only enable the
  entry/exit control if it is allowed to be set to 1.

- Read/write from the guest_ia32_perf_global_ctrl vmcs field only if it
  exists.

vmcs_promote.asm:
Restore host cr8 on promote


connojd:
@rianquinn ok if I merge this?

PR #650, opened 2018-03-20
Add save_state and bitmanip changes for the VIC

connojd:
- Replaces a reserved field in the save_state with an entry
  for the VIC from the extended APIs
- Mark bitmanip functions as constexpr


connojd:
@rianquinn what are your thoughts on the appveyor issue?

PR #658, opened 2018-04-02
Install test_support for extension unit-testing

connojd:
The extended APIs and future extensions will need the intrinsic
stubs defined in test_support.h.

Move test_support.h to bfvmm/include/support/arch/intel_x64
and add install command for CMake.


connojd:
@rianquinn this should be good to go

PR #661, opened 2018-04-13
Add changes for xAPIC in extended APIs

connojd:
- Add MTRR intrinsics
- Include only needed headers into intrinsics unit tests rather than intrinsics.h. This is in support of the MTRR additions, which add a reference to `_cpuid_edx`. By including intrinsics.h, unrelated tests have `_cpud_edx` as an undefined symbol.
- Add move-assignment operator in bfvmm::x64::unique_map_ptr


connojd:
@JaredWright I renamed to '4k', so I think this should be good to go

PR #663, opened 2018-04-16
CTest command line arguments

JaredWright:
Added the ability to specify command line arguments to be passed from
CTest to unit tests


JaredWright:
This change is needed by the eapis EPT unit tests (will open a PR shortly) so that you can specify a command line argument to suppress ASan leak sanitizer errors for unit test executables.

PR #664, opened 2018-04-18
Added: EFI launcher to load Bareflank in early boot

ainfosec-henselb:
This is written mostly in C with a bunch of EFI defines and types and not linked against normal bareflank libraries and can't really resemble bfvmm/eapis code.  But I'm sure it still needs a bunch of refactoring since I've become too familiar with it.


bfdriver/src/platform/efi/.gitignore
```diff
@@ -0,0 +1,5 @@
+src/*.o
```
JaredWright: Can we move these the top-level .gitignore with the rest of the driver exclusions?

bfdriver/src/platform/efi/modules/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
JaredWright: Same comment, move to top-level gitignore

scripts/cmake/config/default.cmake
```diff
@@ -436,6 +436,13 @@ add_config(
     DESCRIPTION "Build unit test components"
 )
 
+add_config(
+    CONFIG_NAME ENABLE_BUILD_EFI
+    CONFIG_TYPE BOOL
+    DEFAULT_VAL ON
```
JaredWright: I think we need to default this to OFF until EFI support becomes a little more stable

scripts/cmake/config/default.cmake
```diff
@@ -746,6 +753,21 @@ set(HIPPOMOCKS_URL_MD5 "6a0928dfee03fbf4c12c36219c696bae"
     "Hippomocks URL MD5 hash"
 )
 
+add_config(
```
JaredWright: Change the add_config macro for GNUEFI_URL and GNUEFI_URL_MD5 to match the other variable definitions of URLs and MD5s just above

scripts/cmake/depends/gnuefi.cmake
```diff
@@ -0,0 +1,43 @@
...
+    message(STATUS "Including dependency: gnuefi")
+
+    download_dependency(
+        gnuefi
+        URL         ${GNUEFI_URL}
+        URL_MD5     ${GNUEFI_URL_MD5}
+    )
+
+    add_dependency(
+        gnuefi userspace
```
JaredWright: How does this work? GNU efi is built for the userspace prefix, but depends on stuff in the VMM prefix?

scripts/cmake/targets.cmake
```diff
@@ -109,6 +109,25 @@ if(NOT WIN32)
...
 endif()
 
+# ------------------------------------------------------------------------------
+# EFI
+# ------------------------------------------------------------------------------
+
+if(NOT WIN32 AND ENABLE_BUILD_EFI)
+
+    add_custom_target(efi_build
+        DEPENDS eapis_efi_${VMM_PREFIX}
```
JaredWright: I dislike that this depends on build targets in the extended apis. Nothing in the base hypervisor should depend on anything in an extension.

bfdriver/src/platform/efi/include/bfmodule.h
```diff
@@ -0,0 +1,53 @@
...
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#ifndef HYP_MODULES_H
+#define HYP_MODULES_H
+
+#include "bfefi.h"
+#include "common.h"
+
```
connojd: Add comments  that describe what the purpose of this file is, and how it integrates with other pieces of your work

bfdriver/src/platform/efi/include/gnuefi/MpService.h
```diff
@@ -0,0 +1,620 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#ifndef _MP_SERVICE_PROTOCOL_H_
+#define _MP_SERVICE_PROTOCOL_H_
+
```
connojd: Please include a reference to the section(s) in the spec from which these defines are derived. Something simple like "Chapter 23, section 4" would be sufficient

bfdriver/src/platform/efi/include/gnuefi/extras.h
```diff
@@ -0,0 +1,68 @@
...
+    /// and are reserved by the UEFI specification for future growth.
+    ///
+    UINT32                           Attributes;
+    ///
+    /// Length in bytes of the FilePathList. OptionalData starts at offset
+    /// sizeof(UINT32) + sizeof(UINT16) + StrSize(Description) + FilePathListLength
+    /// of the EFI_LOAD_OPTION descriptor.
+    ///
+    UINT16                           FilePathListLength;
+    ///
```
connojd: If the fields below are unused, remove these comments

bfdriver/src/platform/efi/include/gnuefi/extras.h
```diff
@@ -0,0 +1,68 @@
...
+    /// that is passed to the loaded image. If the field is zero bytes long, a
+    /// NULL pointer is passed to the loaded image. The number of bytes in
+    /// OptionalData can be computed by subtracting the starting offset of
+    /// OptionalData from total size in bytes of the EFI_LOAD_OPTION.
+    ///
+    // UINT8                         OptionalData[];
+} EFI_LOAD_OPTION;
+#pragma pack()
+
+#endif
```
connojd: Make sure to remove all trailing whitespace and that unix-style newlines are used

bfdriver/src/platform/efi/include/start.h
```diff
@@ -0,0 +1,29 @@
...
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#ifndef BF_START_H
+#define BF_START_H
+
+#include "bfefi.h"
+
+VOID bf_start_hypervisor_on_core(VOID *);
```
connojd: documentation comments

bfdriver/src/platform/efi/src/bfloader.c
```diff
@@ -0,0 +1,77 @@
...
+    add_hypervisor_modules();
+
+    Print(L"Loading modules..\n");
+    int64_t ret = common_load_vmm();
+    if (ret < 0) {
+        Print(L"common_load_vmm returned %a\n", ec_to_str(ret));
+        goto fail;
+    }
+
+    //bf_start_by_interactive();
```
connojd: The commented-out calls should be removed. We could enable the different start modes by a compiler flag or some config argument

bfdriver/src/platform/efi/src/boot.c
```diff
@@ -0,0 +1,98 @@
...
+EFI_STATUS bf_boot_next_by_order()
+{
+    EFI_GUID global_guid = EFI_GLOBAL_VARIABLE;
+
+    EFI_STATUS status;
+    EFI_LOADED_IMAGE *li;
+    status = gBS->HandleProtocol(this_image_h,
+                                 &gEfiLoadedImageProtocolGuid,
+                                 (VOID **)&li);
+    //bf_dump_hex(0,0,li->LoadOptionsSize,li->LoadOptions);
```
connojd: We have a debug switch 'NDEBUG' defined in bfconstants.h. We should use that here or create an EFI-specific debug switch that turns on print statements

bfdriver/src/platform/efi/src/console.c
```diff
@@ -0,0 +1,65 @@
...
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+#include "bfefi.h"
+#include "bfloader.h"
+
+int printf(const char *format, ...)
+{
+
+    if (!format) { return 0; }
```
connojd: use 1TBS style

bfdriver/src/platform/efi/src/console.c
```diff
@@ -0,0 +1,65 @@
...
+    CHAR8 *traveler = (CHAR8 *)format;
+    UINTN counter = 1;
+    while (*traveler != '\0') {
+        traveler++;
+        counter++;
+    }
+
+    CHAR16 *buf = (CHAR16 *)AllocateZeroPool(counter << 1);
+    if (!buf) { return 0; }
```
connojd: 1TBS style

bfdriver/src/platform/efi/src/platform.c
```diff
@@ -0,0 +1,309 @@
...
+ * @return returns the total number of CPUs available to the driver.
+ */
+int64_t
+platform_num_cpus(void)
+{
+    EFI_STATUS status = EFI_NOT_FOUND;
+    UINTN N = 0;
+    UINTN NEnabled;
+
+    if (g_mp_services != NULL)
```
connojd: Add 1TBS-style braces

bfdriver/src/platform/efi/src/x86_64.S
```diff
@@ -0,0 +1,98 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+.text
+
+.global _writeeflags
```
connojd: this and the one below should use 'rflags' rather than 'eflags'

scripts/cmake/targets.cmake
connojd: Agreed. Also, when I tried to build with no eapis enabled, I got a 'missing eapis_efi_x86_64-vmm-elf' make target error

bfdriver/src/platform/efi/Makefile
```diff
@@ -0,0 +1,66 @@
+
+
+TOP			:= 	$(shell pwd)
```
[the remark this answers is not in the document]
brendank310: Make requires the use of actual tabs without some compatibility breaking extensions (might be acceptable to use those extensions, it's been around for a few years now.)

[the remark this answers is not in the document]
JaredWright: Ah, good call. Ignore my comment and keep the tabs

bfdriver/src/platform/efi/gen_modules.sh
```diff
@@ -0,0 +1,50 @@
...
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+
+modularize()
+{
+    cp $1 modules/
+    cd modules
+    filename=$(basename "$1")
+    xxd -i "$filename" > ${filename}.c
```
brendank310: Dumping all the binary bits into a source file seems a little heavy handed, is it possible to use EFI_LOAD_FILE protocol instead?

ainfosec-henselb:
Rian had said he wanted bareflank compiled into the EFI binary vs. fetching it at boot time.  This was months ago and I'm not sure if that's still the case but if it is yes this should probably be done with linker script instead.

________________________________

scripts/cmake/depends/gnuefi.cmake
ainfosec-henselb: This is a very good question and should have a little RFC.  gnu-efi uses some C library headers which it normally gets from the system you're building things on.  For build portability and consistency, I have it use the ones we use in our project.  They're only installed to the vmm prefix though, and because gnu-efi isn't part of the vmm, using headers from the vmm prefix seems like bad architecture.  Merely copying some clib headers from vmm to userspace prefix doesn't isn't any better though.  The project already relies on the system having various tools and libraries present, so maybe this attempt at portability should be scrapped.  Need some input

scripts/cmake/targets.cmake
ainfosec-henselb: ENABLE_BUILD_EFI should default to off so hypervisor builds standalone normally, but enabling ENABLE_BUILD_EFI makes it so some eapis code is necessary.  Attempting to use the unmodified bfvmm in EFI would be pretty useless.  So, bfdriver EFI loader essentially relies on eapis.  Should the EFI loader itself be moved to extended apis?  extended_apis is only necessary if ENABLE_BUILD_EFI is on.

bfdriver/src/platform/efi/Make.efi
```diff
@@ -0,0 +1,37 @@
...
+GCCVERSION   := $(shell $(CC) -dumpversion | cut -f1 -d.)
+GCCMINOR     := $(shell $(CC) -dumpversion | cut -f2 -d.)
+USING_CLANG  := $(shell $(CC) -v 2>&1 | grep -q 'clang version' && echo clang)
+
+# Rely on GCC MS ABI support?
+GCCNEWENOUGH := $(shell ( [ $(GCCVERSION) -gt "4" ]           \
+                          || ( [ $(GCCVERSION) -eq "4" ]      \
+                               && [ $(GCCMINOR) -ge "7" ] ) ) \
+                        && echo 1)
+ifeq ($(GCCNEWENOUGH),1)
```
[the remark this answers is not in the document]
ainfosec-henselb: I hadn't checked before, but clang dropped in surprisingly easily

scripts/cmake/targets.cmake
ainfosec-henselb: It uses little eapis code and the extension could possibly live in bfdriver, just using base intrinsics instead of the eapis abstractions.  Just another possibility.

PR #666, opened 2018-04-26
Add capstone dependency

connojd:
- The xAPIC emulation requires the VMM to map in and decode
  the guest instruction to be emulated.

- Add capstone to depends to support disassembling in the VMM

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ if(NOT WIN32)
 endif()
 
 include_dependency(newlib)
+include_dependency(capstone)
 include_dependency(llvm)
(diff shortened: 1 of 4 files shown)


rianquinn:
Is this needed for the base hypervisor, or can it be added to the EAPIs because dependencies can be added to any repo

connojd:
I dont mind putting it in the ext. apis, although I will need some help to get it building I think. This patch adds the two .a .so libs to the LD_FLAGS in the cmake macro, which probably is the Wrong Way, especially if they are moved the eapis

rianquinn:
No longer needed

PR #670, opened 2018-05-17
bfdriver: Change windows and EFI to MIT license

connojd:


rianquinn:
@connojd Why exactly is AppVeyor failing? Do we know? If not, I can try to compile on Windows and see what happens. I know that the bfdriver doesn't compile on Windows at the moment, but that should not be affecting AppVeyor.

Issue #671, opened 2018-05-18
[BUG] make test is deprecated


rianquinn:
CMake no longer supports creating a custom make test target and our use of this feature is about to be deprecated.

rianquinn:
Fixed

PR #672, opened 2018-05-18
Fix Tidy / Format and move to C++17

rianquinn:
This patch fixes the following issues:
- fixes bugs with clang tidy not working for a diff
- fixes bugs with astyle format not working for a diff
- moves to C++17
- adds JSON to the VMM again for extensions
- removes support for clang 3.8 and 3.9
- removes hardcoded clang versions
- adds upstream versions tidy and format

[ISSUE]: https://github.com/Bareflank/hypervisor/issues/671

Signed-off-by: Rian Quinn <rianquinn@gmail.com>


scripts/cmake/macros.cmake
```diff
@@ -1262,11 +1296,9 @@ endmacro(set_bfm_vmm)
 #     is not in the same directory, allowing you to pass a source file with
 #     a directory. Nomrally FILENAME should still match the filename being
 #     used.
-# @param CMD_LINE_ARGS Command line arguements to be passed to the test
-#     executable by ctest
 #
 function(do_test FILENAME)
-    set(multiVal DEFINES DEPENDS SOURCES CMD_LINE_ARGS)
```
connojd: why is this being removed? I think we should wait until the memory leaks are fixed in the eapis.

rianquinn: I added this back, but I will say that this is a dirty hack. A PR should not be accepted if it has memory leaks, so the tests should fail. At a minimum, if the code is merged, it should be merged with a failure so that it haunts us until its fixed as a memory leak is really bad (way worse than a missing unit test)

rianquinn: Oh.... and I have no idea how it was removed in the first place. Some sort of merge error

rianquinn:
@connojd This is good to go.

PR #678, opened 2018-05-31
MTRR Intrinsics

connojd:
Add MTRR intrinsics for use in extensions


bfintrinsics/include/arch/intel_x64/mtrr.h
```diff
@@ -3818,6 +2571,134 @@ namespace fix4k_F8000
...
+        case write_through:
+        case write_protected:
+        case write_back:
+            return true;
+        default:
+            return false;
+    }
+}
+
+struct variable_range
```
JaredWright: This struct implementation feels kinda awkward to me, and doesn't seem like it belongs in the intrinsics. What do you think about refactoring this into helper functions that would live in the EAPIs? Something that could be used like:

```
mtrr::is_valid(base, mask, type, pas);
mtrr::contains_address(base, mask, addr);
```

It just kinda seems to me like this struct is only used for validating facts about MTRRs, and doesn't really need to have a constructor/internal state to accomplish that. Are there other use cases for this struct that I'm missing?

bfintrinsics/include/arch/intel_x64/mtrr.h
```diff
@@ -37,1254 +37,82 @@ constexpr const auto write_through = 0x04ULL;
...
-
-    inline void dump(int level, std::string *msg = nullptr)
-    {
-        type::dump(level, msg);
-        fe::dump(level, msg);
-        e::dump(level, msg);
-    }
-}
-
-namespace physbase0
```
JaredWright: I really like condensing these all down into just "physbase", nice!

bfintrinsics/include/arch/intel_x64/mtrr.h
connojd: I encapsulated them in a class because there are some many requirements on their form; I thought it would be easier for the end user.

That being said, I have no problem changing to the standard representation. The user of the interface will just have to know what they're doing

connojd: Another reason I just remembered..the eapis currently read the variable ranges and put them in the structs, which are held in a vector. This enables users to overload bool operator() to make the struct Compare-aware. So rather than iterate from start to end as the current eapis does now, they could sorted in a sorted data structure and the search would be logarithmic rather than linear.  

Using only namespace helpers precludes this sort of thing. In practice, I think it would be fine though, because their are only a max of 256 variable MTRRs, and actual implementations only that i've tested so far only have 10

JaredWright: That does make sense. With that in mind, I could see it both ways (using helpers, or using a struct). However whichever one we go with, I still think it belongs in the extended apis.

JaredWright:
@connojd This looks ready to me, do you have any more outstanding changes? I'll merge once you confirm.

connojd:
Good to go

PR #680, opened 2018-06-01
VT-d Intrinsics

JaredWright:
Added numerous intrinsics for manipulating Intel VT-d control structures

* root entries
* extended root entries
* context entries
* pasid entries
* pasid state entries
* first-level paging entries
* second-level paging entries
* fault records
* interrupt remapping table entries
* posted interrupt descriptors

### DIFF
--- a/bfintrinsics/include/arch/intel_x64/vtd/context_entry.h
+++ b/bfintrinsics/include/arch/intel_x64/vtd/context_entry.h
@@ -1,0 +1,174 @@
+//
+// Bareflank Hypervisor
+// Copyright (C) 2018 Assured Information Security, Inc.
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2.1 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+//
+
+#ifndef VTD_CONTEXT_ENTRY_H
+#define VTD_CONTEXT_ENTRY_H
+
+#include <stdint.h>
+#include <bfgsl.h>
+#include <bfbitmanip.h>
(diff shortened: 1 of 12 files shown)


JaredWright:
Heads up: I'll probably need a good amount of feedback on the PR before its ready. As a good starting point, check out bfintrinsics/include/arch/intel_x64/vtd/context_entry.h. Most of the VT-d control structures are larger than 64 bits, so I came up with some creative ```using value_type = ``` statements to account for this.

Also in full disclosure, I generated all of these intrinsics using Shoulder. With that in mind, if there are any issues I'm guessing that they will show up identically across every file.

connojd:
I've done an initial skim and have thought of a couple points:

If the spec introduces an initialism, we should prefer that to the spelled out name i.e ctp rather than context-table pointer.  This is shorter and is more consistent with the other paging-like namespaces (pdpte, pte, ept, etc).

We have to be careful with how we access the VT-d register space. I like the struct data[size] paradigm, but we will need to create or at least consider a read/write interface to access the actual registers appropriately..I'm mainly thinking of an optimizing compiler that replaces an 8-uint84_t posted interrupt descriptor read with one vector instruction which would be bad.

I'll give a more thorough look soon and come back with any input

JaredWright:
1) I'm definitely ok with turning everything into an initialism! That would really help out things like "first level paging structure" and "interrupt remapping structure entry for posted interrupts".

2) I didn't include any intrinsics yet to access the IOMMU registers that control VT-d, just the table structures that those registers will point to. I was thinking that the register access intrinsics would be different enough to require special attention in a separate PR later.

connojd:
:+1:

JaredWright:
So I choose not to use *all* initialisms introduced in the manual, because almost every field of each structure has an initialism, and if we use them all we'll end up with the intrinsics looking like this:

```
vtd::ecte::smep::enable(val);
vtd::ecte::slptp::set(val, 0xf00dbeef);
```

That looks almost obfuscated to me, I'm gunna have a hard time remembering what all those stand for.

Instead, I'm thinking we should adopt initials for each of the table/entry names, and then use the long names for the fields:

```
vtd::ecte::supervisor_mode_execute_protection::enable(val);
vtd::ecte::second_level_page_table_pointer::set(val, 0xf00dbeef);
```

What do you think?

rianquinn:
I hate, hate, hate initialisms and fully support the use of long names.

JaredWright:
The naming currently does not match exactly what is in the manual. To accomplish that, we would need to do the exact opposite of my suggestion above, and make the intrinsics look like this:

```
vtd::extended_context_table_entry::smep::enable(val);
vtd::extended_context_table_entry::slptptr::set(val, 0xf00dbeef);
```

As per our offline discussion last week, I will update all of the intrinsics to that style by the end of today.

PR #681, opened 2018-06-01
Return values from bit intrinsics

connojd:


connojd:
Yeah I'm not sure how the CI didn't catch that. It was until I ran appveyor for the extended apis that the error showed up

PR #686, opened 2018-06-12
CR3 Update

rianquinn:
This patch provides an improved CR3 implementation


rianquinn:
@connojd This should be good to go. Thoughts?

PR #699, opened 2018-06-29
EPT paging intrinsics

JaredWright:
1) Moved ept paging intrinsics from eapis to base
2) Updated ept intrinsics to match the most recent cr3 paging intrinsics


JaredWright:
@rianquinn Here is half of the updates for making the EAPIs EPT implementation similar to the new CR3 one. It is not quite identical, but is very similar.

Question: Do you know if AMD's implementation of EPT equivalent is identical? I put this under the intel_x64 namespace because I wasn't sure.

I have the other half working too (the EPT mmap class), but its going to cause a lot of collisions with the PR that @connojd has open in the EAPIs right now. I think it will be easiest to get that pull request completed first, and then I can rebase my changes on top.

Issue #702, opened 2018-07-27
[BUG]: VMM does not handle NMIs


connojd:
The hypervisor does not install an NMI handler. This is an issue, because the VMM has no way of masking NMIs, and normal, non-buggy guests may program interrupts to be delivered via the NMI pin; Recent versions of Linux send the LVT's PMI (for frequency scaling etc) with the NMI delivery mode. 

If the VMM is running when the CPU receives an NMI, it is delivered through the VMM's IDT - but Bareflank doesn't have one, so a triple fault occurs. This brings the core down, and prompts the NMI watchdog on another core to do an INIT, which the base also doesn't have a handler for.

The more time spent in the VMM, the more likely this bug is to occur. I've been able to reliably reproduce it by running a small program that continuously calls cpuid and by changing the system work load, thereby encouraging cpufreq to send PMIs to modify the frequency.

connojd:
Fixed, see #703

PR #703, opened 2018-07-30
Fix for #702 [BUG]: VMM does not handle NMIs

connojd:
Initialize the host IDT with a default exception handler and an NMI handler.
Note that interrupt handlers are not installed, as those are taken care
of in the extended apis.

  - The non-NMI handlers print out the CPU state at the time of the
    exception and halt the CPU.

  - The NMI handler enables NMI-window exiting. This ensures that an NMI
    can occur on any exit, anywhere the VMM has an IDT, and the NMI will
    be injected as soon as the window opens in the guest.

VT-x requires NMI-exiting and virtuall-NMIs controls to be set when
NMI-window exiting is set, so two exit handlers are added; one for
NMI exits, the other for NMI-window exits. The base exit handler
registers these during construction.

Add unit tests for new code


bfvmm/include/test/intrinsics.h
```diff
@@ -41,6 +41,40 @@ uint16_t g_tr;
 ::x64::gdt_reg::reg_t g_gdtr{};
 ::x64::idt_reg::reg_t g_idtr{};
 
+extern "C" void _handle_nmi(void) noexcept {}
```
rianquinn: no needed

connojd: @rianquinn this is a stub for unit tests

PR #720, opened 2018-09-27
Removed phys_iommu from VT-d intrinsics and fixed missing header in windows driver

JaredWright:


rianquinn:
@JaredWright This is defined incorrectly:
```cpp
static uintptr_t base_addr = 0;
```

It should be:
```cpp
inline uintptr_t base_addr = 0;
```

Defining it as static will cause each file that includes this header to create a static version of this variable that is not shared between compilation units. Declaring it as inline is a new C++17 feature that will create a single instance of this variable that is shared.

PR #727, opened 2019-01-11
Boxy related changes

rianquinn:
This is a huge PR that pulls in a lot of the work that we have been doing over the past couple of months on Linux guest support.

---
This change is [Reviewable](https://reviewable.io/reviews/bareflank/hypervisor/727)


rianquinn:
Sounds good

PR #735, opened 2019-01-27
Post EAPIs merge refactor

JaredWright:
1) Added new bareflank extension mechanism
2) Removed hardware manipulation from all constructors
3) Changed vmm entry code to call all constructors on the bootstrap core
4) Initial implementation of delegator-based exit handlers


JaredWright:
@rianquinn When you get a chance, can you see if this PR breaks your latest version of Boxy? I'm hoping it shouldn't break it at all.

This PR might be ready, pending any comments from review. I can leave some comments on it tomorrow to explain some of the major points, since this is a pretty huge change set.

example/bfvmm_main/src/main.cpp
```diff
@@ -0,0 +1,43 @@
+//
```
JaredWright: I moved bfvmm_main to be an example project. This provides an example of all three override points that this PR adds for extension writters. Details:

* vmm_init gets called on a single core only, and the domain that gets passed as an argument has fully contructed vcpus for each core on the user's system. One key note about this function is that it does NOT run in vmx-root.
* vmm_main gets called on each core in vmx-root as the first thing that happens immediately after the vcpu is loaded (triggered by a special CPUID leaf, just like how it worked before).
* vmm_fini gets called in vmx-root just before destruction is about to happen.

When you take all three of these functions together, there should be no more need for vcpu inheritance of any kind.

CMakeLists.txt
```diff
@@ -151,20 +151,22 @@ if(ENABLE_BUILD_VMM)
...
-
-    add_subproject(
-        integration vmm
-        DEPENDS bfvmm
-        SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/bfvmm/integration/
+        SOURCE_DIR ${SOURCE_EXAMPLE_DIR}/bfvmm_main/src
     )
 endif()
 
+# if(ENABLE_BUILD_TEST AND ENABLE_BUILD_VMM)
```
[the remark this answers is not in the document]
rianquinn: I am fine with working on the integration tests one by one so long as the test_all.cpp test is maintained and works so that we can see an example of how its going to work, and the test_all.cpp is a good one to ensure that the APIs are working.

bfdriver/src/common.c
```diff
@@ -425,9 +425,11 @@ common_start_vmm(void)
             break;
     }
 
+    platform_call_vmm_on_core(0, BF_REQUEST_VMM_INIT, (uint64_t)platform_num_cpus(), 0);
+
     for (cpuid = 0, g_num_cpus_started = 0; cpuid < platform_num_cpus(); cpuid++) {
         ret = platform_call_vmm_on_core(
-                  (uint64_t)cpuid, BF_REQUEST_VMM_INIT, (uint64_t)cpuid, 0);
+                  (uint64_t)cpuid, BF_REQUEST_VMM_RUN, (uint64_t)cpuid, 0);
```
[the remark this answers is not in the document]
rianquinn: - object construction cannot occur on a single core as that will mess up the thread objects. Each vCPU needs to be created on the core it will run on to make sure that the thread objects are initialized properly. 
- same thing, we cannot create the vCPUs all at the same time. They have to be done one by one.
- A default domain can be created as a global object without issue. The domain's constructor is called during global construction on Core 0. 

In other words, no additional function is needed

bfdummy/src/main/dummy_main.cpp
```diff
@@ -57,10 +57,10 @@
 #define REQUEST_SET_RSDP_RETURN ENTRY_ERROR_UNKNOWN
 #endif
 
-#ifndef REQUEST_VMM_INIT_FAILS
-#define REQUEST_VMM_INIT_RETURN ENTRY_SUCCESS
+#ifndef REQUEST_VMM_RUN_FAILS
+#define REQUEST_VMM_RUN_RETURN ENTRY_SUCCESS
 #else
-#define REQUEST_VMM_INIT_RETURN ENTRY_ERROR_UNKNOWN
+#define REQUEST_VMM_RUN_RETURN ENTRY_ERROR_UNKNOWN
```
rianquinn: See above, these should all be removed

bfsdk/include/bfgsl.h
```diff
@@ -158,9 +158,7 @@ auto memset(span<DstElementType, DstExtent> dst, T val)
 
 #else
 
-#ifdef NEED_STD_LITE
-#include <bfstd.h>
```
rianquinn: Why was this changed?

bfsdk/include/bftypes.h
```diff
@@ -138,4 +138,22 @@ typedef INT_PTR intptr_t;
 #define PRId64 "lld"
 #endif
 
+/* -------------------------------------------------------------------------- */
```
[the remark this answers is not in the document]
rianquinn: If a header is needed, lets create one. bytypes is included in the drivers, EFI, etc... which have no idea what the VMM is. This is not the right place for this.

bfsdk/include/vmm.h
```diff
@@ -0,0 +1,43 @@
+//
```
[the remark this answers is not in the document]
rianquinn: Same thing. The SDK is not VMM specific. This type of stuff should be in the VMM, not in the SDK. The SDK is only for headers that are shared across all of the bf*** folders.

bfvmm/include/domain/domain.h
```diff
@@ -0,0 +1,214 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+#ifndef BFVMM_DOMAIN_H
+#define BFVMM_DOMAIN_H
+
+#include <bftypes.h>
+#include <bfobject.h>
+#include <bfgsl.h>
+#include <bfarch.h>
```
rianquinn: ?

bfarch is only needed for an arch file. This is a generic file, so no bfarch is needed.

bfvmm/include/domain/domain.h
```diff
@@ -0,0 +1,214 @@
...
+    /// vcpus
+    ///
+    /// Return the list of vcpus that belong to this domain
+    ///
+    /// @expects
+    /// @ensures
+    ///
+    /// @return List of vcpus
+    ///
+    std::list<vcpu_t>
```
rianquinn: Define an alias for this so that it can be changed later on.

bfvmm/include/domain/domain.h
```diff
@@ -0,0 +1,214 @@
...
+    {
+        return m_vcpus;
+    }
+
+private:
+
+    /// @cond
+
+    domainid_type m_id;
+    std::list<vcpu_t> m_vcpus;
```
rianquinn: Why a list and not a deque?

We should consider removing the vcpu_manager and moving it to the domain. Storing the vCPUs in the domain creates a weak reference which means we would need to ensure that a domain ALWAYs out lives a vCPU which this does not. In other words, you could end up with invalid weak references which breaks the ownership model. Technically, a design like this would require the use of shared pointers, not unique pointers, which is not the direction I would like to take. Instead, we should discuss who owns the vCPUs. If the domain owns the vCPUs, the management of the vCPUs should be in the domain. If the domain doesn't own the vCPUs, it should not be storing a list of them. You should ask the owner. Otherwise, the lifetime of the vCPU is hard to reason about which is a C thing, not a modern C++ thing

bfvmm/include/hve/arch/intel_x64/delegator/cpuid.h
```diff
@@ -0,0 +1,157 @@
...
+    /// Set this to true if your handler returns true and has already
+    /// advanced the guest's instruction pointer.
+    ///
+    /// default: false
+    ///
+    bool ignore_advance;
+};
+
+using delegate_t = delegate<bool(vcpu_t, info_t &)>;
+#define handler(function) delegate_t::create<function>()
```
rianquinn: if these are generic, they should be moved to their own header. If these are not, the name is to generic and will collide with other headers that define the same thing.

Is there a way to do this using templates and auto?

bfvmm/include/hve/arch/intel_x64/delegator/cpuid.h
```diff
@@ -0,0 +1,157 @@
...
+#define stateful_handler(class, object, member) delegate_t::create<class, &class::member>(object)
```
rianquinn: stateful is not a great name for this. You can have a class that doesn't manage state. I would call it a class_handler

bfvmm/include/hve/arch/intel_x64/delegator/unhandled_exit.h
```diff
@@ -0,0 +1,72 @@
...
+    /// Destructor
+    ///
+    /// @expects none
+    /// @ensures none
+    ///
+    ~delegator() = default;
+
+    /// Handle
+    ///
+    /// Handle an unhandled vmexit using registered handlers
```
connojd: Can we use a different name for this? Like 'default' or 'fallback'?

bfvmm/include/hve/arch/intel_x64/domain.h
```diff
@@ -0,0 +1,111 @@
...
+/// @expects
+/// @ensures
+///
+/// @return returns a pointer to the domain being queried or throws
+///     and exception.
+///
+#define get_domain(a) \
+    g_dm->get<bfvmm::intel_x64::domain *>(a, "invalid domainid: " __FILE__)
+
+#define domain_cast(a) domain_t(dynamic_cast<bfvmm::intel_x64::domain *>(a.get()))
```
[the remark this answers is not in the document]
rianquinn: Nice catch. dynamic_cast is an expensive runtime cast that should only be used when you do not know what the incoming type is.

bfvmm/include/hve/arch/intel_x64/exit_handler.h
```diff
@@ -202,71 +147,52 @@ class EXPORT_HVE exit_handler
...
-    void write_control_state();
+    // These members provide the legacy exit handler mechanism, and will be
+    // deprecated slowly
+    std::list<init_handler_delegate_t> m_exit_handlers;
+    std::list<init_handler_delegate_t> m_init_handlers;
+    std::list<fini_handler_delegate_t> m_fini_handlers;
+    std::array<std::list<handler_delegate_t>, 128> m_exit_handlers_array;
+
+    // These members will provide the new exit handler mechanism
+    std::array<handler_delegate_t, 128> m_exit_delegates;
```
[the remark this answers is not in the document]
rianquinn: why do we have two arrays?

bfdriver/src/common.c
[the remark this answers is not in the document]
JaredWright: Can you point me to where thread object initialization happens?

bfvmm/include/hve/arch/intel_x64/exit_handler.h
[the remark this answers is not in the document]
JaredWright: One array is for the "legacy" mechanism (i.e. the current EAPIS style handlers) and the second array is for the new delegator based mechanism. They both exists so that we can slowly deprecate one exit reason at a time.

bfvmm/include/hve/arch/intel_x64/delegator/cpuid.h
JaredWright: These may or may not be generic, I don't think we can tell until all the handers are moved over.

I agree there is probably a better way to do this, I'd like to push off solving that problem until a future PR when we start deprecating old handlers.

bfvmm/include/domain/domain.h
JaredWright: Agreed, not the best implementation, but its a start. I'd rather see a separate PR dedicated to figuring out ownership of vCPUs, because I think we'll run into some complexities there.

bfsdk/include/bfgsl.h
JaredWright: Has to do with defining the vcpu_t types in the bftypes header. Can remove after moving those to a separate header.

bfvmm/include/hve/arch/intel_x64/delegator/unhandled_exit.h
JaredWright: Sure, I like fallback

bfvmm/include/hve/arch/intel_x64/delegator/cpuid.h
JaredWright: I think the name makes sense. Why would you use a class-based handler if the handler didn't need to track state?  class_handler kinda confuses me, because a class isn't a handler, a member function of a class is a handler.

JaredWright: Removed for now, just left the "handler" macro, and moved it to the new generic bfvmm.h header

JaredWright:
I pushed two commits that get rid of the changes to common.c, plus removes the whole concept of a domain. I made changes to entry.cpp that resemble the changes you proposed in your PR set, and am working through one last issue with that right now.

When that gets fixed, I'm guessing that squashing my commits into one will be the easiest the merge?

rianquinn:
yeah, I would squash them all

JaredWright:
Squashed. Hopefully Appveyor and Travis are running smooth today. If you spot any critical errors that I missed, I can get them fixed ASAP today.

I ran the test_all.cpp integration test and the regular bfvmm_main a million times on my Gigabyte test box, and it was stable on my environment.

I did notice that some of the existing integration tests don't quite compile with these changes (looks like really minor things), so I left them commented out for now instead of trying to redo them all. My original thinking was to reintroduce each integration test as each API gets "C-iffied", so I'll leave it to future PRs to introduce those back one at a time.

PR #737, opened 2019-02-02
New End User Interface

rianquinn:
This patch provides a new end user interface for creating extensions
that is more "C" like in nature. The big change that was needed to
support this patch was the removal of the gsl::not_null from delegate
types which is not needed since the APIs can ensure that the vcpus are
always valid making the APIs more accessible to C programmers.

We also removed dead code in the entry.cpp file that was added when the
RSDP code was added but never removed, and we added two additional
init functions for the new C interface


bfvmm/integration/arch/intel_x64/test_all_inheritance.cpp
```diff
@@ -0,0 +1,98 @@
...
+void
+init()
+{
+    ept::identity_map(g_guest_map, MAX_PHYS_ADDR);
+}
+
+namespace test
+{
+
+class vcpu : public bfvmm::intel_x64::vcpu
```
rianquinn: The important thing to notice here is that the original vCPU extension model still works. In fact, all of the unit tests and integration tests still work fine. The only change that will be needed by downstream users is to remove the gsl::not_null which includes Boxy which also still works fine.

bfvmm/integration/arch/intel_x64/test_all.cpp
```diff
@@ -26,82 +26,30 @@
 //     from the EPT map.
 //
 
-#include <bfcallonce.h>
+#include <vmm.h>
```
rianquinn: One change that I made is the `using namespace bfvmm::intel_x64` was moved to the include so that you don't have to do this manually.

bfvmm/integration/arch/intel_x64/test_all.cpp
```diff
@@ -26,82 +26,30 @@
...
-// -----------------------------------------------------------------------------
-
-namespace test
-{
-
-bfn::once_flag flag;
 ept::mmap g_guest_map;
 
-class vcpu : public bfvmm::intel_x64::vcpu
+void
```
rianquinn: Notice the removal of the domain object. I do not believe this is needed for the base as it doesn't do anything without guest support. Also, notice that we don't return bool. If you have an error, throw.

bfvmm/integration/arch/intel_x64/test_all.cpp
```diff
@@ -26,82 +26,30 @@
...
+test_handler(
+    vcpu_t *vcpu, external_interrupt_handler::info_t &info)
 {
+    vcpu->queue_external_interrupt(info.vector);
+    return true;
+}
 
-std::unique_ptr<vcpu>
-vcpu_factory::make(vcpuid::type vcpuid, bfobject *obj)
+void
```
rianquinn: same thing here. no bool. throw if you have an error

bfvmm/integration/arch/intel_x64/test_all_inheritance.cpp
rianquinn: Oh, and near the top we use the init() so using a hybrid approach also works with this design. I could have implemented this without any changes to entry.cpp (that is what I had working last night before bed), but I wanted the old style to work with the new style and the best way to do that was to move the once_call to the entry. The logic is identical so we are not introducing possible issues with threading. All this does is move the once_call from the vcpu constructor to after the vcpu is fully constructed.

bfvmm/src/entry/entry.cpp
```diff
@@ -28,26 +28,31 @@
...
 #include <memory_manager/memory_manager.h>
 
-#include <intrinsics.h>
+static bfn::once_flag g_init_flag;
 
-extern "C" int64_t
-private_init(void)
-{ return ENTRY_SUCCESS; }
+void
+WEAK_SYM init()
```
rianquinn: so the only changes to this file are:
- delete the unused code from the old hyperkernel/rsdp changes. This reverts back to the version of entry.cpp that we had prior to the hyperkernel v1 which is good because I never liked that approach. This only removed "seems". It doesn't change the control-flow
- Add in overridable init functions. Originally I put this in the vCPU class which worked great but providing support for a hybrid approach got me thinking that full construction of the vCPU might be helpful so that you can create your own vCPU, and get a version of it that is fully constructed so the code was moved to here instead.

rianquinn: The important thing to notice here is that the control-flow is identical to the previous version of this file, and no mods were made to common.c in the driver (since the control-flow is the same). Construction and execution all occur on the proper cores.

rianquinn:
OBE

PR #739, opened 2019-02-02
Move MSR and CR code from exit handler

rianquinn:
This code moves the MSR and CR code from the exit handler into
the already existing vmexit handlers that exist. This will need
to be done to the test of the exit handlers, but this is a
good first start at fixing this code now that the EAPIs is
merged into the base hypervisor


rianquinn:
OBE

PR #742, opened 2019-02-25
Rework the VM exit handlers

rianquinn:
This patch addresses several regressions with the previous merge
as well as starts the process of our VM exit rework, which changes
the APIs that VM exit handler provide to be more flexible now that
Boxy has uncovered several shortcomings with the current API
structure.


bfvmm/include/hve/arch/intel_x64/exit_handler.h
```diff
@@ -133,66 +145,19 @@ class EXPORT_HVE exit_handler
...
 public:
 
-    /// @cond
-
-    cpuid::delegator *cpuid_delegator() const
-    { return m_cpuid_delegator.get(); }
-
-    nmi::delegator *nmi_delegator() const
-    { return m_nmi_delegator.get(); }
+    static void handle(
```
JaredWright: Why does this take an exit_handler as an argument instead of just using ```this```?

bfvmm/include/hve/arch/intel_x64/vcpu.h
```diff
@@ -598,6 +623,65 @@ class EXPORT_HVE vcpu : public bfvmm::vcpu
     ///
     VIRTUAL void enable_monitor_trap_flag();
 
+    //--------------------------------------------------------------------------
+    // Non-Mackable Interrupt Window
```
JaredWright: Mackable->Maskable

bfvmm/include/hve/arch/intel_x64/vmcs.h
```diff
@@ -205,75 +186,14 @@ class EXPORT_HVE vmcs
...
-    /// @return returns the VMCS's io bitmap b
-    ///
-    gsl::not_null<uint8_t *> io_bitmap_b() const
-    { return m_io_bitmap_b.get(); }
+    VIRTUAL bool check() const noexcept;
 
 private:
 
-    page_ptr<save_state_t> m_save_state;
+    vcpu *m_vcpu;
```
JaredWright: 👎  Why did this circular relationship make it back into the code again?

bfvmm/include/hve/arch/intel_x64/vmexit/cpuid.h
```diff
@@ -149,53 +128,109 @@ class EXPORT_HVE cpuid_handler
...
+    /// report the vm exit as unhandled, generating an exception. This is
+    /// mostly useful for guest vCPUs that wish to halt() the vCPU if a
+    /// register is accessed that is not explicitly being emulated.
+    ///
+    /// @expects
+    /// @ensures
     ///
-    void set_default_handler(const ::handler_delegate_t &d);
+    void enable_whitelisting() noexcept
+    { m_whitelist = true; }
```
JaredWright: Need a way to disable this too, or one way operation?

CMakeLists.txt
```diff
@@ -151,15 +151,12 @@ if(ENABLE_BUILD_VMM)
         DEPENDS bfunwind
     )
 
-    vmm_extension(
-        bfvmm_main
+    add_subproject(
+        bfvmm_main vmm
         DEPENDS bfvmm
-        DEPENDS bfintrinsics
-        SOURCE_DIR ${SOURCE_EXAMPLE_DIR}/bfvmm_main/src
```
JaredWright: Are we planning to put the example directory back?

bfvmm/include/hve/arch/intel_x64/vmexit/nmi_window.h
```diff
@@ -44,58 +41,87 @@
...
-    /// @expects none
-    /// @ensures none
+    /// Inject an NMI on the next VM entry.
     ///
-    /// @param vcpu The vcpu the vmexit occurred on
+    /// @expects
+    /// @ensures
     ///
-    /// @return True if the vmexit was successfully handled, false otherwise
+    /// @param vector the vector to inject into the guest
```
connojd: delete

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
```diff
@@ -45,6 +45,12 @@
 #define EXPORT_HVE
 #endif
 
+// -----------------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------------
+
+void emulate_wrmsr(::x64::msrs::field_type msr, ::x64::msrs::value_type val);
```
connojd: Can we rename this or take it out of the public header? It does the exact opposite of what emulate means now

bfvmm/include/hve/arch/intel_x64/vmexit/control_register.h
```diff
@@ -136,7 +84,32 @@ class EXPORT_HVE control_register_handler
 
 public:
 
-    /// Add Write CR0 Handler
+    /// Add wrcr0 Handler
+    ///
+    /// Adds a VM exit handler. When a VM exit occurs, the registered
```
connojd: Not a fan of the copy-pasted comments. Can consolidate this info to a single place, then add exit-specific comments as needed here (and elsewhere)?

bfvmm/include/hve/arch/intel_x64/vmexit/cpuid.h
```diff
@@ -149,53 +128,109 @@ class EXPORT_HVE cpuid_handler
...
+    /// enabled, this behavior is disabled, and the base implementation will
+    /// report the vm exit as unhandled, generating an exception. This is
+    /// mostly useful for guest vCPUs that wish to halt() the vCPU if a
+    /// register is accessed that is not explicitly being emulated.
+    ///
+    /// @expects
+    /// @ensures
     ///
-    void set_default_handler(const ::handler_delegate_t &d);
+    void enable_whitelisting() noexcept
```
connojd: What is it that we're whitelisting?

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
```diff
@@ -275,6 +276,8 @@ class EXPORT_HVE wrmsr_handler
     /// @endcond
 };
 
+using wrmsr_handler_delegate_t = wrmsr_handler::handler_delegate_t;
```
connojd: Why are there specific types for each handler? I thought all handlers would have the same type with this change. Or maybe that's yet to come?

bfvmm/src/entry/entry.cpp
```diff
@@ -28,31 +28,35 @@
...
-#include "entry/entry.h"
-#include "bfvmm.h"
-
 static bfn::once_flag g_init_flag;
 
-extern "C" int64_t
-private_init(void)
-{ return ENTRY_SUCCESS; }
+void
+WEAK_SYM global_init()
```
connojd: Is there documentation explaining these new extension functions?

bfvmm/src/hve/arch/intel_x64/vcpu.cpp
```diff
@@ -920,27 +1338,35 @@ vcpu::set_idt_limit(uint64_t val) noexcept
...
 vcpu::cr0() const noexcept
-{ return vmcs_n::guest_cr0::get(); }
+{ return vmcs_n::cr0_read_shadow::get(); }
 
 void
 vcpu::set_cr0(uint64_t val) noexcept
-{ vmcs_n::guest_cr0::set(val); }
+{
+    vmcs_n::cr0_read_shadow::set(val);
+    vmcs_n::guest_cr0::set(val |= m_global_state->ia32_vmx_cr0_fixed0);
```
connojd: Drop the '='

bfvmm/src/hve/arch/intel_x64/vcpu.cpp
```diff
@@ -920,27 +1338,35 @@ vcpu::set_idt_limit(uint64_t val) noexcept
...
 vcpu::cr4() const noexcept
-{ return vmcs_n::guest_cr4::get(); }
+{ return vmcs_n::cr4_read_shadow::get(); }
 
 void
 vcpu::set_cr4(uint64_t val) noexcept
-{ vmcs_n::guest_cr4::set(val); }
+{
+    vmcs_n::cr4_read_shadow::set(val);
+    vmcs_n::guest_cr4::set(val |= m_global_state->ia32_vmx_cr4_fixed0);
```
connojd: Drop the '='

bfvmm/src/hve/arch/intel_x64/vmexit/control_register.cpp
```diff
@@ -306,19 +330,56 @@ void
...
+    vcpu->set_cr4(vcpu->gr1());
 }
 
 // -----------------------------------------------------------------------------
 // Handlers
 // -----------------------------------------------------------------------------
 
 bool
-control_register_handler::handle(vcpu_t vcpu)
+control_register_handler::handle(vcpu *vcpu)
```
connojd: Why are some args wrapped with gsl::not_null but not others?

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
```diff
@@ -21,47 +21,144 @@
...
+    /// example, any memory mapped resources such as ACPI or VT-d need to be
+    /// initalized using the VMM's CR3, and not the hosts.
+    ///
+
+    vcpu_init_root(vcpu);
+    return vcpu->advance();
+}
+
+static bool
+handle_cpuid_0x4BF00011(vcpu *vcpu)
```
connojd: Can we start a list of constexprs or defines of the bareflank-specific cpuids rather than using the hex value?

bfvmm/include/hve/arch/intel_x64/exit_handler.h
rianquinn: Because it is a static function so there is no this. Technically RDI stores the "this" point if we wanted to use a member function and remove "static", but its really hard to read code like that as you really have to know a lot about the System V calling convention to understand how that works.

bfvmm/include/hve/arch/intel_x64/vcpu.h
rianquinn: fixed

bfvmm/include/hve/arch/intel_x64/vmcs.h
rianquinn: Both the local and global state structures for the vCPU are stored in the vCPU as these are not VMCS objects. The launch() function needs a pointer to the vCPU to get access to some of this information (register data really). Although it is circular, it is technical "weak" which is fine. We could remove all weak references, but it would really mess up the function calls as you would have to pass every object to every function for all classes, which doesn't make any sense. 

Most of the circular references were removed, and we pass the vCPU around where it makes sense. In this case, every function needs the vCPU and since its a "weak" reference, it is guideline compliant, safe, and free of any construction ordering issues. IMO, this provides a nice balance/compromise, and later on, we could always continue to remove more of these over time as needed.

bfvmm/include/hve/arch/intel_x64/vmexit/cpuid.h
rianquinn: This is a one-way operation.

CMakeLists.txt
rianquinn: Absolutely. All of the existing examples will be ported to this new design and included in the base repo. Once this is done, all of the example repos will be deprecated.

bfvmm/include/hve/arch/intel_x64/vmexit/nmi_window.h
rianquinn: Nice catch. This was removed in the second PR as doxygen found this too. 
Fixed

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
rianquinn: Correct, this will be removed from the public header when I make the change as the "execute" function will be used instead. I need to leave it here for now for Boxy until I finish this API change.

bfvmm/include/hve/arch/intel_x64/vmexit/control_register.h
rianquinn: I'd like to do both. It is really important that the details of each of these are well documented. We should also provide this same documentation in our examples. More documentation is a good thing IMO, this way no matter what file you land on, you will get the needed information to properly implement a handler/emulator

bfvmm/include/hve/arch/intel_x64/vmexit/cpuid.h
rianquinn: CPUID. Once you turn this on, you have to produce a whitelist of valid CPUID leaves. If you have not registered an emulator for a leaf, it will be reported as unhandled. If this is not enabled, the implementation will fallback to registered handlers. This is needed to ensure that guest vCPUs explicitly emulate all hardware.

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
rianquinn: This will be removed once I complete all of the APIs. This was added for Boxy in the mean time.

bfvmm/src/entry/entry.cpp
rianquinn: No... we need a place to put that. The entry code is not the right place. A future PR absolutely will have to document all of these functions. Any volunteers?

bfvmm/src/hve/arch/intel_x64/vcpu.cpp
rianquinn: nice catch

rianquinn: fixed

bfvmm/src/hve/arch/intel_x64/vcpu.cpp
rianquinn: fixed

bfvmm/src/hve/arch/intel_x64/vmexit/control_register.cpp
rianquinn: All APIs that are "inputs" have gsl::not_null when the contract expects that. All "output" functions, which is mostly the hanlders do not as the user can assume that the APIs will not provide a null pointer. Technically... gsl::not_null should be used for output as this dictates "ensures" and not "expects", but in a previous converstation you me and Connor agreed that this was relatively pedantic, and made the delegates too verbose.

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
rianquinn: yes. Let's do that in another PR, but I agree... for the BF specific leaves, this should be English, not in hex. For hardware that actually exists I prefer the hex as you have to know the hex anyways (the English version for Intel doesn't mean much).

rianquinn:
@JaredWright thanks a ton for the review. I think I have addressed all of your comments. Huge help.

rianquinn:
@connojd Thanks a ton as well. Also, a huge help

bfvmm/include/hve/arch/intel_x64/vmcs.h
JaredWright: My opinion is that all of these weak vcpu references are causing a real headache in terms of separating concerns. If a vcpu needs to know how to control a vmcs, and a vmcs needs to know how to control a vcpu, then they might as well be the same object.

bfvmm/include/hve/arch/intel_x64/exit_handler.h
JaredWright: I didn't even notice that this was a static function, thats even more confusing to me.

If there is going to be a single instance of an entry point for all vmexits, I would think that separating it completely from this class would be a better solution.

bfvmm/include/hve/arch/intel_x64/vmcs.h
rianquinn: Not really for a couple of reasons:
- The vCPU provides the public API that an object needs. In C, this is the same thing as including a header in a circular manner which is done on a regular basis for the same reason. The circular reference is built into Intel's spec and will be there one way or another. The IO instruction code is a great example of this. There is no way to prevent the circular reference as Intel designed it that way. 
- Qt is a great example of why a "weak" reference is an important tool. It breaks the code up into logical components that are easier to test on their own instead of all in one. Yes, putting all of the code into the vCPU would remove the weak references, but it would create a giant blob of code that would be hard to test. Over time, our goal should be to move code from the vCPU into their own classes, even if a weak reference is needed to do it, for the sole purpose of ensuring that a class only has a single purpose/job, which is a common design pattern of C++. 

As for separating, there is no need to do so. A vCPU cannot function without the VMCS and vice versa, and there really isn't a need in the code to find a way to separate them. With respect to APIs and testing, nothing is affected negatively so IMO, there really is no headache here as there is really nothing to do/fix. For a lack of better words, I fail to see the problem it creates. Maybe a better way to say it is, there is a lot for us to fix/work on, this is very low on the priority list for me to fix since it is not getting in our way, and attempting to fix this now clutters our interfaces and pulls us from other more important things.

bfvmm/include/hve/arch/intel_x64/exit_handler.h
rianquinn: I am not opposed to that. I would take that a step further and make it a C extern so that the asm code that references this function doesn't have any mangling in it. I will upload that change.

rianquinn: Ok, so I remember why we did this. It was a friend function of an extern "C" function issue (and me simply not wanting to figure out the black magic to make that work). 

Fixed

JaredWright: Nice! I really like this solution, because it gets rid of the name-mangling in the exit handler entry assembly. 👍

bfvmm/include/hve/arch/intel_x64/vmcs.h
JaredWright: > * but it would create a giant blob of code that would be hard to test

Correct, this is what the vCPU already is. I just disagree that these week references are the best way to solve this problem

> * Over time, our goal should be to move code from the vCPU into their own classes

Agreed, and thats the most important part. As long as thats the end goal, lets leave this weak reference for now

JaredWright:
I'm happy with this PR, so I vote to merge it once CI finishes! @connojd?

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
connojd: :+1:

bfvmm/include/hve/arch/intel_x64/vmexit/control_register.h
connojd: I'm all for documenting in every file, but if something is copy-pasted more than once within that file, I think it should just be put once at the top or something

bfvmm/include/hve/arch/intel_x64/vmexit/cpuid.h
connojd: So if a leaf is whitelisted, that means that you don't need an emulator for it?

bfvmm/include/hve/arch/intel_x64/vmexit/wrmsr.h
connojd: :+1:

bfvmm/src/entry/entry.cpp
connojd: I can do that. I've started some documentation on my cmake branch

bfvmm/src/hve/arch/intel_x64/vmexit/control_register.cpp
connojd: Ok sgtm I just couldn't remember they were for inputs

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
connojd: >  this should be English, not in hex.
:+1: 

> For hardware that actually exists I prefer the hex as you have to know the hex anyways
Can you give an example of this? You said hardware; what is the actual scope you're referring to? What specific registers/devices/etc?

connojd:
@rianquinn I had just a couple follow up questions, but lgtm

JaredWright:
Awesome. @rianquinn Do you want to squash these before we click merge?

PR #748, opened 2019-03-12
Fix detection of the EPT switching capability

antonburtsev:
Barefank was handling the 0x491 MSR (aka Capability Reporting Register of VM-function Controls) incorrectly. The 0x491 is a simple MSR that doesn't have separate parts for 0- and 1-settings, so we don't have to shift the mask by 32 to check if a specific VM-function, e.g., EPT switching, is supported. See section A.11 of the Intel SDM.


rianquinn:
@antonburtsev Thanks a ton for the PR. I will merge once the tests pass.

PR #750, opened 2019-03-16
CMake changes

connojd:
Introduce a new target-centric cmake design


.codecov.yml
```diff
@@ -44,6 +44,7 @@ ignore:
   - "bfintrinsics/**"
   - "bfm/ioctl/arch/**"
   - "bfvmm/include/test/**"
+  - "bfdriver/**"
```
rianquinn: why was this added? I'd like to work at reducing these. Not adding to them.

.doxygen.txt
```diff
@@ -43,7 +43,7 @@ PROJECT_BRIEF = "lightweight, extendable hypervisor written in C++ with support
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
-INPUT = bfdriver/include bfelf_loader/include bfm/include bfsdk/include bfvmm/include
```
rianquinn: why was this removed?

.travis.yml
```diff
@@ -275,127 +291,3 @@ matrix:
...
-          - clang-6.0
-          - nasm
-    env:
-      - TEST="GCC 8"
-    script:
-      - cmake -DCONFIG=travis_shared ..
-      - make -j3
-
-  #
-  # ld.gold
```
rianquinn: why are all of these tests removed?

.doxygen.txt
connojd: I moved the elf loader to bfsdk

.codecov.yml
connojd: The driver tests arent being built right now because we dropped shared library support

JaredWright:
I think this is HUGE improvement, great work. So far, I've run into one issue: dependencies between top level projects seem to be missing, breaking parallel builds on my environment. Specifically, the bfvmm subproject seems to be missing direct (or transitive) dependencies on newlib, binutils, libcxx, libcxxabi, bfintrinsics, bfsdk. Are you having issues with parallel builds on your environment?

Also, it looks like the next ubuntu release (19.04) comes with cmake 3.13 (woo hoo!) We should keep it on our radar to update the README next month to make that our minimum supported versions, since this PR moves to cmake > 3.13

JaredWright:
Changing top-level CMakeLists.txt:139 from ```add_subproject(bfvmm vmm DEPENDS bfunwind)``` to ``` add_subproject(bfvmm vmm DEPENDS bfunwind bfintrinsics) ``` fixed the dependency issues on my environment

PR #756, opened 2019-03-25
Boxy Suspend Support Changes

rianquinn:
This patch provides API changed need to support
suspend/resume in Boxy with guest support. Specifically this
patch does the following:
- Adds some additional handlers for launch/resume
- Ensures that clearing a VMCS will force a relaunch
- Adds a VMCall interface to the driver that can be used to
  ensure that vmcall is only called when the hypervisor is
  running

### DIFF
--- a/bfvmm/include/hve/arch/intel_x64/vcpu.h
+++ b/bfvmm/include/hve/arch/intel_x64/vcpu.h
@@ -115,9 +115,13 @@ public:
 
     /// Run Delegate
     ///
-    /// Provides the base implementation for starting the vCPU. This delegate
-    /// does not "resume" a vCPU as the base implementation does not support
-    /// guest VMs.
+    /// Provides the base implementation for starting the vCPU. This will
+    /// either launch the vCPU or resume it (if it has already been launched).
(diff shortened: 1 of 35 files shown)


bfvmm/include/hve/arch/intel_x64/vcpu.h
```diff
@@ -126,16 +130,39 @@ class EXPORT_HVE vcpu : public bfvmm::vcpu
     ///
     VIRTUAL void run_delegate(bfobject *obj);
 
-    /// Halt Delegate
+    /// Add Launch Delegate
+    ///
+    /// Adds a launch delegate to the VCPU. The delegates are added to a queue and
+    /// executed in FIFO order. All delegates are executed unless an exception
```
connojd: Is there a reason this if FIFO and not LIFO like the other delegate lists?

rianquinn: The same reason the run, hlt, init and fini delegates don't. In this case, there is some core functionality that has to run, so the delegates are not overridding like the VMexit handlers.

PR #758, opened 2019-04-04
Cpuid sdk

JaredWright:
This patch series introduces the concept of SDK functions to the base hypervisor, using CPUID as a guinea-pig feature. The idea behind SDK functions is two-fold:

1) Attempt to decouple an extension writer's needs from the historically turbulent vCPU class. For example:

```
namespace bfvmm::intel_x64::cpuid
{

// The goal is to keep this interface stable...
void add_handler(vcpu *vcpu, leaf_t leaf, handler_delegate_t handler) 
{
    // ... while allowing this interface to remain dynamic
    vcpu->add_cpuid_handler(leaf, handler); 
}

}
```
2) Translate operations that expose a vCPU's implementation details into more intuitive "english":

```
namespace bfvmm::intel_x64::cpuid
{

// This function will strive to present intuitive meaning...
leaf_t get_leaf(vcpu *vcpu)
{
    // ... while this function can use the less intuitive implementation details of a vCPU
    return vcpu->gr1();
}

}
```

Eventually, we should work to create SDK functions that cover all common operations that an extension writer needs to accomplish (MSRs, control registers, ept, etc). Also, I think that this should eventually be moved to the top level of the project to ```hypervisor/sdk``` to make it clear that it is a separate layer from ```hypervisor/bfvmm``` (can't move it now, because we already have something else called ```hypervisor/bfsdk''')

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -41,10 +41,11 @@ parsers:
       macro: no
 
 ignore:
-  - "bfintrinsics/**"
-  - "bfm/ioctl/arch/**"
-  - "bfvmm/include/test/**"
-  - "bfdriver/**"
+  - "bfintrinsics/**"           # Not useful to unit test
+  - "bfvmm/include/test/**"     # Don't calculate coverage on unit test code
+  - "bfdriver/**"               # Not possible to unit test (OS drivers)
+  - "bfvmm/include/sdk/**"      # Covered with integration tests
+  - "bfvmm/src/sdk/**"          # Covered with integration tests
 
   # Remove Me
   - "bfelf_loader/**"
--- a/bfvmm/CMakeLists.txt
+++ b/bfvmm/CMakeLists.txt
@@ -73,6 +73,7 @@ endif()
 
 if(PREFIX STREQUAL vmm)
     add_subdirectory(src/entry)
+    add_subdirectory(src/sdk)
     add_dependencies(bfvmm bfvmm_entry)
     target_link_libraries(bfvmm INTERFACE
         --whole-archive
@@ -85,6 +86,7 @@ if(PREFIX STREQUAL vmm)
         bfvmm_vcpu
         bfvmm_memory_manager
         bfvmm_debug
+        bfvmm_sdk
         vmm::bfintrinsics
         ${CMAKE_INSTALL_PREFIX}/lib/libc++.a
         ${CMAKE_INSTALL_PREFIX}/lib/libc++abi.a
--- a/bfvmm/include/vmm.h
+++ b/bfvmm/include/vmm.h
@@ -65,6 +65,8 @@
 #include "hve/arch/intel_x64/vcpu.h"
 #include "hve/arch/intel_x64/vmx.h"
 #include "hve/arch/intel_x64/vpid.h"
+
+#include "sdk/cpuid.h"
 #endif
 
 #ifdef BF_X64
@@ -81,6 +83,9 @@
 
 #ifdef BF_INTEL_X64
 using namespace bfvmm::intel_x64;
+
+#define handler_delegate(function) handler_delegate_t::create<function>()
+
 #endif
 
 #endif
--- a/bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
+++ b/bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
@@ -179,9 +179,6 @@ cpuid_handler::add_emulator(
 void
 cpuid_handler::execute(gsl::not_null<vcpu *> vcpu)
 {
-    vcpu->set_gr1(vcpu->rax());
-    vcpu->set_gr2(vcpu->rcx());
-
     auto [rax, rbx, rcx, rdx] =
         ::x64::cpuid::get(
             gsl::narrow_cast<::x64::cpuid::field_type>(vcpu->rax()),
@@ -227,21 +224,20 @@ execute_emulators(vcpu *vcpu, const std::list<handler_delegate_t> &emulators)
 bool
 cpuid_handler::handle(vcpu *vcpu)
 {
-    const auto &emulators =
-        m_emulators.find(vcpu->rax());
+    vcpu->set_gr1(vcpu->rax());
+    vcpu->set_gr2(vcpu->rcx());
+
+    const auto &emulators = m_emulators.find(vcpu->rax());
 
     if (emulators != m_emulators.end()) {
         return execute_emulators(vcpu, emulators->second);
     }
 
     if (m_whitelist) {
-        vcpu->set_gr1(vcpu->rax());
-        vcpu->set_gr2(vcpu->rcx());
         return false;
     }
 
-    const auto &handlers =
-        m_handlers.find(vcpu->rax());
+    const auto &handlers = m_handlers.find(vcpu->rax());
 
     this->execute(vcpu);
 
--- a/bfvmm/src/sdk/CMakeLists.txt
+++ b/bfvmm/src/sdk/CMakeLists.txt
@@ -1,0 +1,35 @@
+#
+# Copyright (C) 2019 Assured Information Security, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+add_library(bfvmm_sdk)
+
+target_link_libraries(bfvmm_sdk PUBLIC ${PREFIX}::bfroot)
+
+target_include_directories(bfvmm_sdk PUBLIC
+    $<${BUILD_INCLUDE}:${PROJECT_ROOT}/include>
+)
+
+target_sources(bfvmm_sdk PRIVATE
+    $<${X64}:cpuid.cpp>
+)
+
+install(TARGETS bfvmm_sdk DESTINATION lib EXPORT bfvmm-${PREFIX}-targets)
+install(DIRECTORY ../../include/sdk/ DESTINATION include/bfvmm/sdk)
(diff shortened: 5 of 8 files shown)


.codecov.yml
```diff
@@ -41,10 +41,11 @@ parsers:
       macro: no
 
 ignore:
-  - "bfintrinsics/**"
-  - "bfm/ioctl/arch/**"
-  - "bfvmm/include/test/**"
-  - "bfdriver/**"
+  - "bfintrinsics/**"           # Not useful to unit test
+  - "bfvmm/include/test/**"     # Don't calculate coverage on unit test code
+  - "bfdriver/**"               # Not possible to unit test (OS drivers)
```
rianquinn: that comment is not true. once the ELF loader is fixed to remove shared library support, we will be able to test this again.

.codecov.yml
```diff
@@ -41,10 +41,11 @@ parsers:
...
+  - "bfvmm/include/sdk/**"      # Covered with integration tests
```
rianquinn: I am fine with this so long as this is header only and only wraps around the existing APIs.

.codecov.yml
```diff
@@ -41,10 +41,11 @@ parsers:
...
+  - "bfvmm/src/sdk/**"          # Covered with integration tests
```
rianquinn: this shouldn't be needed

bfvmm/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ endif()
 
 if(PREFIX STREQUAL vmm)
     add_subdirectory(src/entry)
+    add_subdirectory(src/sdk)
```
rianquinn: remove

bfvmm/CMakeLists.txt
```diff
@@ -85,6 +86,7 @@ if(PREFIX STREQUAL vmm)
         bfvmm_vcpu
         bfvmm_memory_manager
         bfvmm_debug
+        bfvmm_sdk
```
rianquinn: remove

bfvmm/include/vmm.h
```diff
@@ -81,6 +83,9 @@
 
 #ifdef BF_INTEL_X64
 using namespace bfvmm::intel_x64;
+
+#define handler_delegate(function) handler_delegate_t::create<function>()
```
rianquinn: this is going to cause problems because it will collide with the existing code. This is also not helpful because it prevents the use of classes in the delegate. Just use the long form. Its long, but at least it is explicit as to what you are doing.

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
```diff
@@ -227,21 +224,20 @@ execute_emulators(vcpu *vcpu, const std::list<handler_delegate_t> &emulators)
 bool
 cpuid_handler::handle(vcpu *vcpu)
 {
-    const auto &emulators =
-        m_emulators.find(vcpu->rax());
+    vcpu->set_gr1(vcpu->rax());
+    vcpu->set_gr2(vcpu->rcx());
```
rianquinn: this is good. The only thing that this is missing is in the emulation code, we need to set rax, rbx, rcx, and rdx to 0 now that you have fixed this bug.

bfvmm/src/sdk/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+add_library(bfvmm_sdk)
```
rianquinn: remove this

.codecov.yml
JaredWright: I'll let you remove the comment and the exclusion once it becomes possible to do those tests in the future.

.codecov.yml
JaredWright: I don't want to make these header-only for a couple reasons:

1) The point of the SDK functions is to define an interface that hides implementation details. Seperating into a header and source is a logical way to accomplish that separation.

2) When we put implementations in a header file, the files end up looking like [this](https://github.com/Bareflank/hypervisor/blob/e3b641225ead1fee2b2c3bdeb2c32d84e4265f44/bfvmm/include/hve/arch/intel_x64/ept/helpers.h). In my opinion, there's a lot of unnecessary "C++ clutter" in there that distracts from learning how to use the interface.

bfvmm/include/vmm.h
JaredWright: Maybe just use a different name for the macro, like ```handler()```? What does it collide with?

I find ```handler_delegate_t::create<function>()``` not simple, and just ugly. Also, I disagree that this prevents the use of classes in the delegate, you can always use the non-macro syntax if you want it.

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
JaredWright: Which scenario should rax/rbx/rcx/rdx always be zeroed out? Does Intel specify that CPUID returns 0 for all leafs that don't exist?

.codecov.yml
rianquinn: sounds good

.codecov.yml
rianquinn: The problem is, the source approach adds symbols to the runtime linking phase and it prevents inlining, so the compiler cannot optimize out your code. Basically, you are adding additional function calls to every API for the sole purpose of reducing clutter. Also, its not a C++ thing, inlining is done in C as well. 
https://www.kernel.org/doc/local/inline.html

bfvmm/include/vmm.h
rianquinn: handler_delegate is contained in handler_delegate_t (same with handler). Downstream files could end up with collisions. With that said, I am fine with trying to find a meta programming approach to reducing their verbosity as the class version of this can get along quickly with no great way to line break. 

Also... I completely disagree with the function version of this. I think it is very elegant. It reads just like a std::unique_ptr<>. Its the class version that sucks. @connojd and I spent some time to try and make the class version a little less verbose without success but I am open to suggestions that don't leverage macros where possible. We have a bunch of these in Boxy as well that I want to remove as well.

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
rianquinn: Yes, 0 is defined as the return value for CPUID. The emulation code should zero out the registers just before this line:
https://github.com/Bareflank/hypervisor/blob/e3b641225ead1fee2b2c3bdeb2c32d84e4265f44/bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp#L215

.codecov.yml
connojd: Im somewhat torn here, so I'm going to ask some questions first:
> The point of the SDK functions is to define an interface that hides implementation details

@JaredWright  what is the problem(s) that hiding the implementation solves?

> so the compiler cannot optimize out your code

@rianquinn do you have any numbers for this? Are we sure that the lack of optimization has a negative impact for Bareflank specifically?

rianquinn: I don't have any numbers for this specific use-case, but wikipedia claims 10-20%: https://en.wikipedia.org/wiki/Inline_expansion#Effect_on_performance

More importantly, the impact would be a lot higher for this specific case because it's not just about expansion. In this case, the compiler will literally remove the function call wrapper as it is not needed. If you place this code in source, the compiler is forced to mark the code as global (even in the static compilation case), and will be present, meaning these APIs will be adding an additional function call for every API. 

Like I said before, this is not a C++ thing. Here is one of a million examples in the Linux kernel:
https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/apic.h#L102

We should be moving towards inlining, not away from it. Especially if the function is designed to wrap up code that we would normally write ourselves. C++ in a lot of ways is a header-only language. Meta-programming requires it. Just look at libc++ and how it is implemented. The only source files are for things that absolutely need it. I am all for making our APIs more approachable, but not at the expense of turning C++ into C, or hurting performance. 

Clang 5 BTW added modules support. For the VMM at least, we could always add modules support to improve compilation performance. But even with that specific issue, the ship has sailed as the intrinsics are the thing that needs that. These wrappers are just noise.

rianquinn: One way to solve this BTW, is to do what libc++ does:
https://github.com/Bareflank/libcxx/blob/master/include/memory#L26

I might be misunderstanding the objection, but it seems to me that the biggest issue @JaredWright has is the way the code looks. To solve this, we could always do this:
https://github.com/Bareflank/libcxx/blob/master/include/memory#L26

Basicaly, at the start of the file in a comment block, you document all of the functions for Doygen without the implementation, and then the implementation follows without comments. This way, if you are a user, you simply read the comment section like you would an unimplemented header file, but you get the performance benefits of inlining.

JaredWright: @connojd To clarify your question:

> what is the problem(s) that hiding the implementation solves?

The problem that I foresee hiding implementation details (i.e. declaring abstract interfaces) solving is instability between extensions and the base hypervisor. I'm hypothesizing that defining purely abstract interfaces will encourage both us (the library developers) and users (extension writers) to think more in terms of abstract interactions and end user needs, and less in terms of how that need is accomplished at a given moment in time. As @rianquinn points out, lots of things in C++ (including the standard library) are *implemented* as header only. The keyword there is implemented, because even those header files are conforming to a separate abstract specification (the ISO C++ standard). Right now, Bareflank doesn't have this concept of "abstract specification", so I'm asserting that separating headers from implementations will help start us down that path. I have no way to prove if this will be successful, but I'd like to give it a try.

> Are we sure that the lack of optimization has a negative impact for Bareflank specifically?

Agreed on this point. I'm not willing to talk about performance and optimizations at this level (removing one function call) until we have a way to measure those impacts.

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
JaredWright: We might need to think a little more about this one... I thought the purpose of the "emulator" path was to allow extension writers an un-tampered-with path to take over handing of an event. Zeroing out the output registers is slightly tampering with the handling of the event.

bfvmm/include/vmm.h
JaredWright: Connor just demo'ed me a far superior solution to the macro that gives pretty much the same syntax. I'll remove the macro

bfvmm/src/hve/arch/intel_x64/vmexit/cpuid.cpp
rianquinn: huh... good point. this comment got me thinking. In the MSR case, your 100% right. In the event that an MSR is not supported, the proper response is to leave the registers alone and inject a GPF. I'll mark this as resolved. If you emulate, you should zero out the registers manually. It's a bit of extra work, but there really is no clean way to know what to do in each case, so let the delegate figure it out.

bfvmm/include/vmm.h
rianquinn: Yeah. IMO, worst case, we can address this issue using the malloc based delegate which has a really nice and clean interface. We would just have to ensure we know what the malloc size is so that we can create an object_allocator for that allocation size. Since we don't register delegate at runtime, this approach would work too. But yeah, if Connor can get his C++17 type deduction trick to work in the class case, that is the way to do it.

PR #852, opened 2019-04-11
Cpuid interface

JaredWright:
This is a follow-up to PR #758, which (to reiterate here) had the following two goals:

1) Attempt to decouple an extension writer's needs from the historically turbulent vCPU class.
2) Translate operations that expose a vCPU's implementation details into more intuitive "english"

This PR proposes an alternative way to accomplish those two goals, without introducing the previous concept of "SDK functions". To summarize the new approach:

* Logical collections of features provided by a vCPU (using CPUID as the first guinea-pig) will be defined by an abstract [interface](https://github.com/JaredWright/hypervisor/blob/af2d5e7187bc68e2c1c989ba88dfd8c5b2372f19/bfvmm/include/interface/arch/intel_x64/cpuid.h) (a class with nothing besides pure virtual functions and type information)
* Each each function in the interface will represent a user need (i.e. use case) for the vCPU
* The interface will strive *not* to expose any implementation details of how each use case is accomplished. For example,  the ```leaf_t cpuid_vmexit_leaf()``` interface function describes a use case (get the CPUID leaf that caused a VM exit), and is implemented under the hood as a less intuitive implementation detail: ```vcpu->gr1()```
* A vCPU will implement each interface that it would like to provide (for example, the base intel_x64::vcpu will provide a specific implementation of the vcpu_cpuid_interface)

There's a couple specific review points that I have questions about, so I'll leave a self-review here shortly to start some conversation

### DIFF
--- a/bfsdk/include/bftypes.h
+++ b/bfsdk/include/bftypes.h
@@ -84,8 +84,12 @@
 
 #ifdef ENABLE_BUILD_TEST
 #define VIRTUAL virtual
+#define PURE = 0
+#define OVERRIDE override
 #else
 #define VIRTUAL
+#define PURE
+#define OVERRIDE
 #endif
 
 /* -------------------------------------------------------------------------- */
--- a/bfvmm/CMakeLists.txt
+++ b/bfvmm/CMakeLists.txt
@@ -49,6 +49,7 @@ add_subdirectory(src/memory_manager)
 add_subdirectory(src/debug)
 
 install(FILES include/vmm.h DESTINATION include/bfvmm)
+install(DIRECTORY include/interface DESTINATION include/bfvmm)
 
 # ------------------------------------------------------------------------------
 # test
--- a/bfvmm/include/hve/arch/intel_x64/vcpu.h
+++ b/bfvmm/include/hve/arch/intel_x64/vcpu.h
@@ -49,6 +49,8 @@
 #include "vmx.h"
 #include "vpid.h"
 
+#include "../../../interface/arch/intel_x64/cpuid.h"
+
 #include "../x64/unmapper.h"
 
 #include "../../../vcpu/vcpu.h"
@@ -66,7 +68,9 @@ namespace bfvmm::intel_x64
 /// This class provides the base implementation for an Intel based vCPU. For
 /// more information on how a vCPU works, please @see bfvmm::vcpu
 ///
-class vcpu : public bfvmm::vcpu
+class vcpu :
+    public bfvmm::vcpu,
+    public vcpu_cpuid_interface
 {
 
 public:
@@ -347,61 +351,36 @@ public:
     VIRTUAL void execute_wrcr4();
 
     //--------------------------------------------------------------------------
-    // CPUID
+    // vCPU CPUID Interface Implementation
     //--------------------------------------------------------------------------
 
-    /// Add CPUID Handler
-    ///
-    /// Add a delegate to handle a CPUID VM exit. Your handler should always
-    /// return false unless you need to override the default behavior of the
-    /// base hypervisor. For more information, please see the CPUID VM exit
-    /// handler for details.
-    ///
-    /// @expects
-    /// @ensures
-    ///
-    /// @param leaf the leaf to call d on
-    /// @param d the delegate to call on the vm exit
-    ///
-    VIRTUAL void add_cpuid_handler(
-        cpuid_handler::leaf_t leaf, const handler_delegate_t &d);
+    /// @cond
 
-    /// Add Emulate
-    ///
-    /// Emulate the exeuction of the CPUID instruction. For more information
-    /// please see the CPUID VM exit handler for details. Generally speaking,
-    /// you should typically use add_handler() instead.
-    ///
-    /// @expects
-    /// @ensures
-    ///
-    /// @param leaf the leaf to call d on
-    /// @param d the delegate to call on the vm exit
-    ///
-    VIRTUAL void add_cpuid_emulator(
-        cpuid_handler::leaf_t leaf, const handler_delegate_t &d);
+    void cpuid_add_handler(
+        vcpu_cpuid_interface::leaf_t leaf, const handler_delegate_t &d) OVERRIDE
+    { m_cpuid_handler.add_handler(leaf, d); }
 
-    /// Execute CPUID
-    ///
-    /// Executes the CPUID instruction, and populates the vCPU's registers.
-    /// For more information, please see the CPUID VM exit handler's
-    /// documentation.
-    ///
-    /// @expects
-    /// @ensures
-    ///
-    VIRTUAL void execute_cpuid();
+    void cpuid_add_emulator(
+        vcpu_cpuid_interface::leaf_t leaf, const handler_delegate_t &d) OVERRIDE
+    { m_cpuid_handler.add_emulator(leaf, d); }
 
-    /// Enable Whitelisting
-    ///
-    /// Ensures that if a VM exit occurs, that an emulator must be registered
-    /// for the exit. If an emulator is not registered, the VM exit is
-    /// reported as unhandled.
-    ///
-    /// @expects
-    /// @ensures
-    ///
-    VIRTUAL void enable_cpuid_whitelisting() noexcept;
+    void cpuid_execute() OVERRIDE
+    { m_cpuid_handler.execute(this); }
+
+    void cpuid_emulate(uint64_t rax, uint64_t rbx, uint64_t rcx, uint64_t rdx) OVERRIDE {
+        this->set_rax(set_bits(this->rax(), 0x00000000FFFFFFFFULL, rax));
+        this->set_rbx(set_bits(this->rbx(), 0x00000000FFFFFFFFULL, rbx));
+        this->set_rcx(set_bits(this->rcx(), 0x00000000FFFFFFFFULL, rcx));
+        this->set_rdx(set_bits(this->rdx(), 0x00000000FFFFFFFFULL, rdx));
+    }
+
+    leaf_t cpuid_vmexit_leaf() const OVERRIDE
+    { return this->gr1(); }
+
+    leaf_t cpuid_vmexit_subleaf() const OVERRIDE
+    { return this->gr2(); }
+
+    /// @endcond
 
     //--------------------------------------------------------------------------
     // EPT Misconfiguration
--- a/bfvmm/include/interface/arch/intel_x64/cpuid.h
+++ b/bfvmm/include/interface/arch/intel_x64/cpuid.h
@@ -1,0 +1,108 @@
+//
+// Copyright (C) 2019 Assured Information Security, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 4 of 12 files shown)


bfsdk/include/bftypes.h
```diff
@@ -84,8 +84,12 @@
 
 #ifdef ENABLE_BUILD_TEST
 #define VIRTUAL virtual
+#define PURE = 0
+#define OVERRIDE override
```
JaredWright: I'm assuming that this ```#define VIRTUAL``` macro exists as a performance optimization to avoid the overhead of a v-table. For the vcpu interface definitions, I really like the rigid compiler checks that can be accomplished by combining pure virtual functions with the C++11 ```override``` keyword, so what are everyone's thoughts about providing similar macros to accomplish this? For an example of how these new macros are used, see [here](https://github.com/JaredWright/hypervisor/blob/af2d5e7187bc68e2c1c989ba88dfd8c5b2372f19/bfvmm/include/interface/arch/intel_x64/cpuid.h#L49) and [here](https://github.com/JaredWright/hypervisor/blob/af2d5e7187bc68e2c1c989ba88dfd8c5b2372f19/bfvmm/include/hve/arch/intel_x64/vcpu.h#L367)

bfvmm/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ add_subdirectory(src/memory_manager)
 add_subdirectory(src/debug)
 
 install(FILES include/vmm.h DESTINATION include/bfvmm)
+install(DIRECTORY include/interface DESTINATION include/bfvmm)
```
JaredWright: @connojd Is there a better way that I should be installing a directory that contains nothing but header files? Do I need to turn it into an interface library?

bfvmm/include/hve/arch/intel_x64/vcpu.h
```diff
@@ -347,61 +351,36 @@ class vcpu : public bfvmm::vcpu
...
-    /// reported as unhandled.
-    ///
-    /// @expects
-    /// @ensures
-    ///
-    VIRTUAL void enable_cpuid_whitelisting() noexcept;
+    void cpuid_execute() OVERRIDE
+    { m_cpuid_handler.execute(this); }
+
+    void cpuid_emulate(uint64_t rax, uint64_t rbx, uint64_t rcx, uint64_t rdx) OVERRIDE {
```
JaredWright: Astyle is formatting this function really weird on my environment, it won't let me move the opening ```{``` to the next line

bfvmm/include/interface/arch/intel_x64/cpuid.h
```diff
@@ -0,0 +1,108 @@
...
+    using leaf_t = uint64_t;
+
+    /// CPUID Add Handler
+    ///
+    /// Add a handler for a CPUID VM exit caused by the given CPUID leaf.
+    ///
+    /// @param leaf the CPUID leaf register handler @param h for
+    /// @param h the handler to call upon a VM exit
+    ///
+    VIRTUAL void cpuid_add_handler(leaf_t leaf, const handler_delegate_t &h) PURE;
```
JaredWright: I changed the naming convention slightly from what existed previously with the following thoughts:

* the first part of the function identifies the interface that it belongs to: ```cpuid_```
* the rest of the function describes the use case it provides ```add_handler```

I was thinking that code completion tools would be better able to suggest available functionality if we named functions like this. For example, if you started typing vcpu->cpui..., you would see all the functions that belong to the cpuid interface.

Thoughts?

bfvmm/include/test/hve.h
```diff
@@ -172,10 +172,12 @@ setup_vcpu(MockRepository &mocks, ::intel_x64::vmcs::value_type reason = 0)
     mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::execute_rdcr3);
     mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::execute_wrcr3);
     mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::execute_wrcr4);
-    mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::add_cpuid_handler);
-    mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::add_cpuid_emulator);
-    mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::execute_cpuid);
-    mocks.OnCall(vcpu, bfvmm::intel_x64::vcpu::enable_cpuid_whitelisting);
```
JaredWright: I removed the enable_whitelisting function from the cpuid interface per a discussion with Rian the other day. Let me know if I misinterpreted our conversation...

bfsdk/include/bftypes.h
rianquinn: I would remove this change as it is not needed. The VIRTUAL macro is only needed for unit testing. The difference between `VIRTUAL` and `virtual` is that VIRTUAL is used to state that the function can be mocked, but not actually overriden. If it can be overridden, mark it with `virtual` and then any overrides need to be marked with `override`. This is a simple fix for a known issue in C++ with respect to mocking. PURE and OVERRIDE are not needed because those would only be needed if you intend to use the `virtual` version.

bfvmm/include/hve/arch/intel_x64/vcpu.h
```diff
@@ -347,61 +351,36 @@ class vcpu : public bfvmm::vcpu
...
-    /// @ensures
-    ///
-    /// @param leaf the leaf to call d on
-    /// @param d the delegate to call on the vm exit
-    ///
-    VIRTUAL void add_cpuid_emulator(
-        cpuid_handler::leaf_t leaf, const handler_delegate_t &d);
+    void cpuid_add_handler(
+        vcpu_cpuid_interface::leaf_t leaf, const handler_delegate_t &d) OVERRIDE
+    { m_cpuid_handler.add_handler(leaf, d); }
```
rianquinn: these shouldn't be here now that we are moving everything to the interface. The implementation should look like this when we are done.

```
class vcpu : 
    public bfvmm::vcpu,
    public bfvmm::interface::cpuid,
    ....
{ 
public:
    vcpu::vcpu(id) : 
        bfvmm::vcpu(id),
        bfvmm::interface::cpuid(this),
        ....
    { }
};
````

And then each interface actually implements the stable API (using both headers and source)

````
class cpuid
{
public:
    cpuid(vcpu_t *vcpu);

    // stable APIs go here

private:

    // Implementation specific details go here
};
````

Pure virtual base class functions are only needed when you intend to enforce multiple inheritance. In general, however, C++ Concepts are a more prefered mechanism that accomplishes the same thing. In this specific case, the `cpuid` APIs are Intel specific, so there is no reason to provide pure virtual functions as nothing else will inherit this. 

At some point when we attempt to add support for AMD, we will likely end up with some code that is shared in which case it will need to be moved to a non-Intel location, but the APIs and how they are used will still remain stable so long as they adhere to the concept the interface defines. 

Here is an example of a C++ concept (there are many):
http://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/concept/Allocator.html

This interface can be implemented by anything and with whatever private implementation it wants, and the code works without modification so long as the concept is adhered to. Another way to state this is, to create a stable interface, all we have to do is define what the interface is and its requirements. Any archiectural implementation from that point on simply needs to implement this interface to work, and from a stability point of view, modifications to the concept are what must be guarded, knowing that any changes to the concept will effect everyone and so they should only be done if absolutely needed. 

The only issue with concepts today in C++ is that if the code doesn't adhere to the concept, the only way you will know is when you attempt to compile and something doesn't add up which results in weird compiler errors. This is however being address in C++20, and we shouldn't attempt to reinvent this:
https://en.cppreference.com/w/cpp/language/constraints

This is further being expanded in C++20 with the use of contracts:
https://en.cppreference.com/w/cpp/language/attributes/contract

All of this can be used to further define and enforce the stable API once the features are available, and this is the direction the language is taking. Until then, our goal should be to do the same thing the Standard Library does which is to define the concept for each interface and then stick to it. Modifications should only be made if they are absolutely needed to ensure downstream users are not effected. 

One final note about pure virtual functions. The main issue with this approach is it significantly increase the size of the vTable for the object. One main reason that concepts are used over pure virtual interfaces is these interfaces were only meant for functions that are needed by a base implemention. For example, if you have a class that defines an Animal and you then want to create a Dog and a Cat and a function for Talk (the anmial makes a noise), you do not need to inherit and override a base function in the Animal class unless you intend to store only Animals and call the Talk function from your list of Animals. If, however, you intend to use the Cat and Dog classes as they are, you can call the Talk function from any other class through the use of concepts. In fact, there is no need to even define the Animal class in the first place. To take this a step futher, even if you plan to store only Animals and call the Talk function from your base Animal, you still do not need to inherit and override as Type Erasure can be used to implement concepts as well. 

This doesn't mean that inheritance is bad or that it doesn't have its place, it simply means that if concepts can be used instead of inheritance, they should as they prevent the class from ballooning out of propotion in size, and they also prevent the need for a vTable entry which ultanitemly prevents inlining and adds an additional function call when it is not needed to accomplish the same thing. 

P.S. even if a design like this way used, you really should be using "final" instead of "override" unless you intend the APIs to be overriden by the APIs users.

File: bfsdk/include/bftypes.h (hunk 1 above)
Comment (rianquinn) [reply bfsdk/include/bftypes.h]: also, see comments below

File: bfvmm/include/hve/arch/intel_x64/vcpu.h (hunk 3 above)
Comment (rianquinn) [reply bfvmm/include/hve/arch/intel_x64/vcpu.h]: see comments above and below. The issue is you are using a macro which will not be needed when this is fixed.

File: bfvmm/include/interface/arch/intel_x64/cpuid.h (hunk 4 above)
Comment (rianquinn) [reply bfvmm/include/interface/arch/intel_x64/cpuid.h]: also not an issue, see below

File: bfvmm/include/test/hve.h (hunk 5 above)
Comment (rianquinn) [reply bfvmm/include/test/hve.h]: The white list function should be removed. The other functions should not be.

**Comment by rianquinn** [comment]
@JaredWright / @connojd 
https://github.com/rianquinn/hypervisor/commit/f5ea4afd83fe25f77d7525ba51617014e6e6c8e2

Here is my version of this using C++ traits instead of pure virtual inheritance. The main benefits of this approach are:
- there is no additional overhead to this approach as no vTable is needed
- this uses composition and not inheritance
- no need for additional macros
- using inlining, this should perform exactly the same as the code we currently have. 

There are a lot of other additional things that I really like about this approach:
- we can use this to remove the need for a lot of the vCPU resource from having to be accessible in the public API as we can group vmexit handlers up into common API interfaces. 
- the implementation of the additional functions that are not in the vmexit handlers have a good place to live
- unit testing should now be a lot easier to follow, and we should be able to delete the include/test folder when this is done.

File: bfvmm/include/interface/arch/intel_x64/cpuid.h [hunk 7]
```diff
@@ -0,0 +1,108 @@
+//
+// Copyright (C) 2019 Assured Information Security, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
```
connojd: Can we use something other than 'interface' - maybe uapi or just api?

rianquinn: I like uapi.

rianquinn:
@JaredWright / @connojd 

I am working on trying to get this work to actually run (dependency issue with construction) which requires that I port VMX, VMCS, Exit Handler and the state classes to this interface. The other PR that I was working on was a rework of the VMCS APIs to sort out VMCS Migration and other Boxy specific issues so I will be merging these two PRs together since the VMCS stuff that I need is in the other PR.

bfvmm/include/interface/arch/intel_x64/cpuid.h
JaredWright: Are we just talking about the name of the folder? I like uapi too (I'm guessing that is short for user api)

rianquinn: yeah. I already made the change in my code. I like it. Any other comments on the approach? I have the state stuff almost complete. The VMCS is going to take the longest as the API is pretty big (all of the vCPU set/get functions that are for the VMCS). I don't want to invest too much time if we don't like the approach.

connojd: I would prefer uapi to interface and implementation to impl everywhere

connojd:
@rianquinn I left some comments, mainly formatting/naming stuff.

PR #857, opened 2019-04-19
Update clang support

connojd:
- Remove clang 5 travis test
- Add clang 7 and 8 travis tests
- Use '= delete' for the mmap move constructors to silence
  the -Wdefaulted-function-deleted warning introduced with clang 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -251,27 +251,6 @@ matrix:
       - make unittest
 
   #
-  # Clang 5.0
-  #
-  - os: linux
-    addons:
-      apt:
-        sources:
-          - ubuntu-toolchain-r-test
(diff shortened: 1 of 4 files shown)


rianquinn:
@connojd Before I merge this does this address the issue with newlib not using the compiler that is set in the CMAKE command line? I don't think it does. Could you verify that newlib is using the compiler you by ensuring the more than one clang is on the system and that when you select a specific one, newlib is using it?

connojd:
@rianquinn I added a small change for newlib and verified it works in travis and on my system. I built clang from source and newlib respects it

@matt-dees can you verify this change works for your custom clang?

rianquinn:
this is what we have in the toolchain so I think this is the right mod.

Issue #855, opened 2019-04-13
[RFC] Delegate pattern


connojd:
C++17 added [class template argument deduction](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) which enables the compiler to deduce the template arguments of a class. This means that you can call constructors of a template class without explicitly listing the types in angle brackets.

[This](https://github.com/connojd/delegate/blob/master/inheritance/test.cpp#L25) is how the code looks with the new feature. That version uses inheritance to get a common 'delegate<function_signature>' type usable in containers. There are static_delegate and member_delegate subclasses that derive from the abstract delegate class, and both are marked final to help the compiler devirtualize the calls. If we decide to go down this route, I can add subclasses for lambdas too (although I don't think they are really needed).

Obviously this code is easier to read than the existing method, and IMO we should start using it. If others agree, I can work on a more complete version and then measure the performance impact versus the current implementation.

JaredWright:
I 100% agree that these are easier to read, and less verbose that the existing method. 👍 from me on this.

```static_delegate``` does sound a little misleading though (it doesn't have to be a static function, any function will work). Does ```function_delegate``` make sense instead?

I'll let @rianquinn comment on any disadvantages this has over the current approach, because he is probably the most familiar with the current implementation.

**Comment by rianquinn** [comment]
@connojd HUGE fan of this. Here are my notes:
- What is the reason for the subclasses? Technically, we should be able to create a single delegate type with different constructors. Then it can just be called delegate and we are done. 
- I agree with @JaredWright that static is misleading. If we need subclasses, I would just call it delegate and member_delegate
- Based on your implementation and the inclusion of C++17, we should be able to further extend your example such that adding a delegate to a delegate list doesn't actually need you do state the type. For example:

```
bool foo(vcpu_t *vcpu);
vcpu->add_exit_handler({&foo});
```

or 

```
class blah{
public:
    bool foo(vcpu_t *vcpu);
};
blah g_blah;
vcpu->add_exit_handler({&blah::foo, &g_blah});
```

rianquinn:
Oh.. one more thing. We should ensure that only valid delegates can be created. Null delegates should not be allowed. Otherwise, we will have to add a bunch of code to check for null which I would like to avoid.

connojd:
I've created a version using std::function [here](https://github.com/connojd/delegate/blob/master/stdfunc/test.cpp#L33). The downside of this approach is the malloc for member delegates.

[...]

PR #858, opened 2019-04-23
Use CTAD in delegate implementation

connojd:
This series introduces [CTAD](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) with the delegate class. CTAD helps reduce the
verbosity of creating delegates. So far, CTAD works for normal functions, member functions, and lambdas that capture <= 16 bytes. Dynamic allocation for larger objects can be added in a later PR after this initial version has been tested.

See issue #855 for more background discussion on this change

### DIFF
--- a/bfvmm/include/hve/arch/intel_x64/vmexit/ept_violation.h
+++ b/bfvmm/include/hve/arch/intel_x64/vmexit/ept_violation.h
@@ -90,7 +90,7 @@ public:
     /// handlers
     ///
     using handler_delegate_t =
-        delegate<bool(vcpu *, info_t &)>;
+        delegate<bool, vcpu *, info_t &>;
 
     /// Constructor
     ///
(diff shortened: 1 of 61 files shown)


bfvmm/include/hve/arch/intel_x64/vmexit/ept_violation.h
```diff
@@ -90,7 +90,7 @@ class ept_violation_handler
     /// handlers
     ///
     using handler_delegate_t =
-        delegate<bool(vcpu *, info_t &)>;
```
rianquinn: remind me again why this had to change?

connojd: Its doesn't I forgot to change it back to the original. Should be good now

connojd:
@rianquinn this version has null delegates because of the few default handlers. I would like to rework the exit handlers so the check isn't necessary. So operator() is guaranteed to be valid at all times for every delegate, it just may not do anything. I'm sure of the best way to do this in the exit handlers though, since each return value has a specific meaning.

PR #883, opened 2020-01-07
Fix exiting on CR0 write for unrestricted guests

krystian-hebel:
vcpu::set_eptp() didn't disable exiting on writes to CR0. On the first
run (BSP) handlers were set as if there was no support for unrestricted
guests. Default handler for CR0 writes called emulate_ia_32e_mode_switch(),
which doesn't allow for switches between all modes.

Signed-off-by: Krystian Hebel <krystian.hebel@3mdeb.com>

### DIFF
--- a/bfvmm/src/hve/arch/intel_x64/vcpu.cpp
+++ b/bfvmm/src/hve/arch/intel_x64/vcpu.cpp
@@ -891,15 +891,33 @@ vcpu::disable_preemption_timer()
 void
 vcpu::set_eptp(ept::mmap &map)
 {
+    using namespace ::intel_x64::vmcs;
+    using namespace ::intel_x64::cr0;
+
     m_ept_handler.set_eptp(&map);
     m_mmap = &map;
+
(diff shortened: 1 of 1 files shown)


rianquinn:
LGTM +1

andyhhp:
Apologies for reviving a closed but, but how much do you care about Westmere processors?

Unrestricted Guest strictly depends on EPT, but one generation of Westmere processors have EPT and lack Unrestricted Guest.  On these processors, you can use EPT instead of shadow pagetables, but you still may not disable CR0.PG/PE in non-root mode.

The simplest option is to treat Unrestricted Guest as a prerequisite for EPT.  After all, Westmere is an obsolete processor these days.

rianquinn:
Yeah, right now, we don't support those processors. Anyone would be free to implement their own handlers to override our default implementation and provide support as they see fit, but we would prefer to keep the upstream source simple, which requires a more modern implementation of VT-x.

PR #885, opened 2020-01-26
Solid foundation

JaredWright:
I'm guessing that this PR is going to generate a lot of discussion points. Rather than trying to write down all my thoughts about this PR here, an offline discussion might be more productive for a first pass.

To summarize, the PR contains some suggestions for high-level directory and file naming that will support moving the project toward the static interface pattern. This includes some minimal CMake scaffolding that exists purely to demonstrate what the result of building the project would look like.


rianquinn:
Couple of initial impressions:
- I like the top level structure. Its self-documenting and easy to follow. The only thing that it is missing are the .xxx files for the different services that we use. A more complete list of them is in the BSL right now. 
- The cmake sub files would suggest you are using the old system. Cache files (i.e., cache warnings), and even config files are not really needed anymore. Anywhere we have a "set" we could toss that into a "config", but the use of ccmake was dropped a while ago as it doesn't work, so all that would be doing is attempting to place the things you can change on the command line or in your own config/override file. Things like add_dependencies is no longer needed, and there is no "init" subproject as cmake doesn't init anything. 
- WRT to AUTOSAR, every single class, struct, union, and even high level functions need their own file with the exact same name. So if we have a file called "dumper", there is a class in that file call "dumper". The vmmctl is nothing more than an application that executes the IOCTL interface of the VMM's driver. The class breakdown should reflect how we want the IOCTL interface logically separated.  There is no need for command line code, etc... It is literally just, take arguments, send to IOCTL, done. My previous branch on this at least attempted to lay some of this out. It uses Libc++ and argparse stuff which will all change, but the structure I think was pretty close
- I am not really sure what you are trying to show in the VMM folder so I am not going to comment on that yet (I am assuming it is incomplete?) Nothing in there seemed to self-document what its purpose was
- One thing that is missing from all of this is the logical separation of interfaces and implementations. Remember that everything will be a template, so everything is implemented in header files. Where do we put the static interface headers vs the implementation headers?
- One thing that will be needed is a "Map File" class. This will be needed by the vmmctl to map in the VMM and/or VMs. That is an interface we will have to support and extensions should be able to use. 
- The IOCTL interface is another interface that others should be able to use. Specifically, the VMCall instruction has to be executed from the driver so that we can return failure when the VMM is not loaded (or in sleep/resume scenarios). 

Those are my initial thoughts so far. I am sure I will have a lot more as this continues to develop.

JaredWright:
Excellent, it sounds like we're on the same page about pretty much everything. A lot of these bullet points will be addressed once we have a more complete file/directory structure. I'll address each of them incrementally with more PRs to this branch.

PR #886, opened 2020-02-03
Add suggested file structure for vmm/vcpu and vmmctl

JaredWright:
This PR adds some file name suggestions that attempt to establish a pattern for defining interfaces + implementations for vcpu and vmmctl components. It isn't a complete file structure for the whole project yet, but it suggests a pattern for moving forward to other project components. In general, here is what I'd be curious for feedback about (in addition to anything else that jumps out immediately):

* Given the name of each file + path in the directory structure, could you make a reasonable guess as to what will be contained in that file?
* Given only a file's name, could you guess at its path starting from the root of the project given a minimal prior knowledge of the project's structure?
* Is it easy and intuitive to identify a boundary between the project's public interfaces and internal implementations?
* Is it easy and intuitive to see a division between microarchitectures?
* Will the naming support autosar compliance?


connojd:
I have some questions:

1) Why is the arch duplicated in the file name under src? Some of them twice e.g. vmm/src/vcpu/x86_64/x86_64_general_register_x86_64.h and others aren't duplicated at all:  vmm/src/vcpu/x86_64/preemption_timer_intel.h.

2) What will common code look like when it includes arch specific code? Is the build system going to resolve the necessary arch to include? I would really like us adopt something similar to linux where common code just includes <asm/header.h> and the build system takes care to -I the proper arch.

JaredWright:
> Why is the arch duplicated in the file name under src?

The idea there was to use the naming scheme [interface_name].h for interfaces, and [interface_name]_\<implementation_name\>.h for things that implement those interfaces. So the examples you gave would be broken down as:
* (preemption_timer)_\<intel\>.h, where preemption_timer is the interface, and it is an intel-specific implementation. I was thinking of using only "_intel" to denote intel-specific, and using "_x86_64" to denote any x86_64 system (intel or amd).
* [x86_64_general_register]_\<x86_64\>.h, where x86_64_general_register is the interface, and the implementation is shared between both intel and amd. For this particular file, the naming scheme does end up looking pretty ugly!

> What will common code look like when it includes arch specific code?

Common code would not directly include or call arch specific code. Instead, it would use an arch-specific implementation _through_ an interface. The build system and/or component factories (I will introduce that in a later PR) will be responsible for setting up the interface->implementation bindings. That should hopefully give a similar behavior to what you are describing, giving include statements that look like ``` #include <vcpu/interface_name.h> ```

connojd:
Ok that makes sense but to answer one of your original questions, it isn't obvious that the header leaf encodes the interface and implementation of that interface. Perhaps it would be more clear if interface files where in an "interface" directory and vice-versa for implementations?

JaredWright:
So just get rid of "include" and "src" directories altogether, and instead name them "interface" and "implementation"? I could definitely go for that. Especially since everything with this approach ends up in a header file (which many people associate with an include directory, not a src directory).

connojd:
Yeah something like that..I think removing (or renaming) 'src' directory is probably a good idea if there are only going to be .h files

JaredWright:
After some extensive offline discussion, @connojd and I ended up circling back to the following conclusion:
* Keep the "include" directory to contain all of the project's interface definitions (don't rename it to "interface")
* Keep the "src" directory to contain all of the project's implementation details (don't rename it to "implementation")
* Keep the file naming scheme exactly as it stands in this PR.

connojd:
Yes and for anyone wondering, we are trying to attain AUTOSAR/MISRA compliance, which requires that there be one file for each class defined, and the file name must be the same as the class. 

Well what if you put all the *_x86_64 in a `namespace x86_64`? Then you could have each class name be the same as those defined under include, but enclosed in `namespace x86_64` and the file name wouldn't need the redundant '_x86_64'.

Sadly this wont work either, because AUTOSAR only allows one non-global namespace per project

rianquinn:
Comments

- I would not use x86_64 as a common name as that is Intel specific. For AMD is it amd64. Other projects have used x64 as a common name (which I think we currently use). 
- I am not a huge fan of using namespaces for x86_64, amd64 and common because they cannot coexist, which I would like. For example, I see no reason why AMD logic cannot call Intel logic and vice_versa. For this reason, I like the idea of interface_implementation as they can all exist in the same top-level namespace and coexist. 
- I don't think we need a cache warnings file in cmake as that should not be an issue moving forward. 
- I am not sure what init_project and init_subproject are for. We will not have any dependencies anymore (outside of the BSL), which can be managed entirely with EPA, so I am not really sure what a "project" is compared to a subproject, and I am not sure what they do. A subproject, in my mind, is our way of compiling each component in Bareflank as we have to compiler native and cross, and EPA is needed to do that. 
- Not sure what a "constant" is. I am also not sure what a "add_config" would do. Since ccmake is not longer supported, all you need to do is set(VAR VALUE) (inside an ifndef), in a config file and call it a day. If you want to override a value, we simply need to provide a macro to injest a user-config file and source it prior to reading the default config. 
- I don't see anything for the build targets. It might make sense to have a folder for these, with a file for each target, and at the high level, bring it all together. The BSL has a bareflank.cmake which has all of the details (please see my master branch). That file has all of the details, it just needs structure and organization
- We will need macros for adding a test, an example, and will need cmake code for creating the libraries that users will link to. The Standalone CXX project has all of the details for how that works. Specifically, there is code to create the interface library, and then add code to it for things that CMake currently doesn't have a fancy function for. 
- We will also need logic for generating the toolchain file and placing it somewhere that makes sense. This file will need to detect the arch on the system if the arch is not defined. There is a lot of toolchain specific stuff that we will have to generate. Please note that there is also a LOT of crap in there that we do not need now that we are dropping newlib and libcxx, but some of it will still be needed (compiler flags that are not defines are a good example of that). 
- For the vmmctl, I like all of the file names, and I know what all of these are. They, however, need to be called vmmctl_load, vmmctl_run, etc... Otherwise, you are consuming the name "call" and "run" which would then not be able to be used as a function name in other parts of the code. If you are concerned about the code being vmmctl::vmctl_load, etc... we could always just use "bfn" as the top level namespace, with bfn::details being used for private functions, and no other namespaces (besides bsl). 
- For the platform stuff... lets use "lin" for Linux, "win" for Windows, "efi" for UEFI, and "all" for common to help with verbosity. Also.. it might be nice to have the platform first. Specifically, when you have a lot of file open in your editor, a file that is called "parse_arguments_lin" and "parse_arguments_win" might just say "parse_ar..." in the editor, and you wouldn't know which file to click as you are coding. Having the files "lin_parse_arguemnts" would show "lin_ar...." and "win_ar..." which might make it a little easier. Just a thought. 
- For the vmmctl, you are missing a file to the core logic associated with BFM. main.cpp should match interfaces to implementations and then run some sort of "execute" function. Everything else should be implemented in a header file that accepts template parameters for any dependencies it has, including the core logic. This ensures unit testing is easy without the need for mocking. In my test branch "giant_rework2", I called it main.h, but it should be anything. 
- WRT to the vcpu headers, we just need to make sure that we do not anticipate any name collisions. If we think this might be an issue, we could prepend with "vcpu_" to ensure names cannot collide with function names. I like all of the files in there, and they all make sense as well. As stated above, I am not a fan of using x86_64 as that is Intel specific, and it would be nice to see that prepended and not appended.

JaredWright:
Replies for @rianquinn 
* I'd be fine to switch the naming to use "x64" everywhere to mean "common between intel and amd 64-bit"
* All the comments about CMake are outside of scope for this PR, let's figure that out on its own later.
* For your comments about vmmctl interface names, how about using "load_vmm" and "call_vmm" instead? That would give a pattern that looks like ```vmmctl::run_vmm```, which sorta reads like a sentence (i.e. "the vmmctl component runs the vmm")
* I'm not a fan of adopting "lin" and "win" as shorthand notation for linux and windows. As an outsider to the project, if I saw a directory named "lin" and a function called "lin_foo()" I would have no idea what either would be for. I'd also opt to use "uefi" as a name (because efi is technically a different intel-specific specification), and not using any labels/specifiers for common code.
* I'd be fine to switch the naming scheme for all implementation files from [interface]_<implementenation> to  <implementation>_[interface].
* Instead of using a "main.h" interface in vmmctl, how about we add an "application" interface, that contains an execute function?
* WRT the vcpu headers, you're absolutely right that we can't use "run" or "load" for the interface names (those are the only two that will collide with a function name). Instead of prepending "vcpu_" to every interface name, I'd rather come up with an interface that to contain the functions load, unload, run, and hlt. Would it be the "cpu" interface, because those are the operations that bind a vcpu to a cpu? Or the "execute" interface, because all those functions provide execution lifecycle management?

connojd:
* +1 for using x64
* +1 for not using 'lin' and 'win'
* +1 for `vmmctl::run_vmm`

rianquinn:
@JaredWright +1 for all of your comments. 

WRT to the file structure, I like your proposal. Would it make sense for you to focus on the BFM and add the build system/static interfaces? This would give a chance to see the bigger picture, and I can add the guts (as they are mostly done, just need to update to use the BSL instead of Libc++). From there, we could go back to looking at the VMM. The driver logic still needs to be sorted out as well, but I think it makes more sense to get enough in place that would allow us to work with it and see how Helix QAC likes the structure (as it will dictate a lot).

JaredWright:
Dannnnng... looks like I screwed up a rebase + force push, because git was having trouble merging all my commits with just empty files. Since it got so screwed up, I think I'll have to close this PR and address all of the comments here in a new one. Sorry!

PR #887, opened 2020-02-19
Add suggested file structure for vmm/vcpu and vmmctl

JaredWright:


JaredWright:
This PR follows up #886, which got messed up by myself on accident. I believe that this PR addresses all points of conversation between myself, @rianquinn, and @connojd, specifically:

* I changed all "x86_64" names to just "x64"
* I switched the file naming scheme from [interface]\_\<implementenation\> to \<implementation\>\_[interface]
* I switched file extensions from .h to .hpp (because of autosar)
* I did my best to rename interfaces that had a naming collision with functions

This PR won't have a perfect file structure, but it should be a enough of a foundation to start adding some basic interface definitions and CMake for the project. My plan is to start this process for vmmctl in my next PR to this branch.

PR #888, opened 2020-02-22
Add top-level cmake scaffolding

JaredWright:


JaredWright:
Here's an attempt at some CMake that tries to do the following:

* Compile *everything* as a single CMake project (no ExternalProject_Add anywhere)
* Compile from Linux/Windows/MacOS using clang + lld
* Allow you to do configure the build with ```cmake ../hypervisor``` no matter which environment you are on (I found some clever tricks to default the generator to Ninja on Windows, and look for clang as the default compiler, etc.)

Building this PR should produce the following:
* A vmmctl executable that can print "hello world" (build/vmmctl/vmmct)
* The base vmm library (build/vmm)
* An example extension that builds a minimal "hello world" VMM (build/example/hello_world)

Could others try to build this on their own environments and tell me what happens? Also, try to audit the VMM executable with readelf/objdump to see if this is on the right track.

connojd:
I got the following error when running from ubuntu 19.10:
```zsh
❯ make
Scanning dependencies of target vmm
[ 20%] Building ASM-ATT object vmm/CMakeFiles/vmm.dir/src/entry/start.s.obj
[ 20%] Built target vmm
Scanning dependencies of target vmmctl
[ 40%] Building CXX object vmmctl/CMakeFiles/vmmctl.dir/src/main.cpp.obj
[ 60%] Linking CXX executable vmmctl
[ 60%] Built target vmmctl
Scanning dependencies of target hello_world_vmm
[ 80%] Building CXX object example/hello_world/CMakeFiles/hello_world_vmm.dir/src/main.cpp.obj
[100%] Linking CXX executable hello_world_vmm
clang: error: invalid linker name in argument '-fuse-ld=lld'
make[2]: *** [example/hello_world/CMakeFiles/hello_world_vmm.dir/build.make:86: example/hello_world/hello_world_vmm] Error 1
make[1]: *** [CMakeFiles/Makefile2:219: example/hello_world/CMakeFiles/hello_world_vmm.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

JaredWright:
@connojd On Ubuntu, try running a ```sudo apt-get install lld``` first, and then try again.

connojd:
ok that worked! I can test on Windows tomorrow and Ill leave some code comments inline

connojd:
All works from windows too. I tested from normal cmd and the x64 native vs2019 cmd prompt 👍

PreLoad.cmake
```diff
@@ -0,0 +1,50 @@
+# Ninja is the only supported CMake generator on Windows
```
connojd: this preload trick was a good find!

cmake/toolchain/clang.cmake
```diff
@@ -0,0 +1,6 @@
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_C_COMPILER_WORKS 1)
+set(CMAKE_CXX_COMPILER_WORKS 1)
+
+set(CMAKE_C_LINK_EXECUTABLE "\"${CLANG++_BIN}\" ${LD_FLAGS} <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
```
connojd: How will linking for EFI work here? The current [efi](https://github.com/Bareflank/hypervisor/blob/master/scripts/cmake/toolchain/clang_x86_64_efi.cmake) toolchain adds some pre and post commands for creating the EFI application

PreLoad.cmake
JaredWright: I don't even remember how I stumbled across this, but its a pretty handy trick! I tried it on cmake 13, 14, 15, 16 and it works in all versions.

cmake/toolchain/clang.cmake
JaredWright: For those linker flags in the EFI toolchain, I think we could convince CMake to add them in the right order using target properties.

However, this also brings up a larger issue of how to build/link anything that implements a loader for the vmm (windows drivers, linux drivers, arm bootloaders, etc). Most of those likely won't support building with clang/lld, so we might have to use ExternalProject to build those components. I'll try to add a follow up commit to this PR that handles building those as well.

PR #892, opened 2020-03-29
Solid foundation

JaredWright:
This PR provides 2 things:

1) Adds dependencies for bsl and pal to the build system using CMake's FetchContent module. This approach has some serious pros/cons against using ExternalProject + submodules/subtrees. I'll add some comments in the code below to highlight the major ones.

2) Provides an autosar compliant, SOLID conforming strategy for defining the top-level components in vmmctl. The pattern goes like this:

* Interfaces are defined in vmmctl/include
* Implementations are all defined in vmmctl/src
* The binding between the two happens in numerous main functions under vmmctl/main, which serve a purpose pretty close to an abstract factory pattern (each main function acts kinda like a concrete factory). CMake chooses an appropriate main function based on your build configuration.
* A class called vmmctl::app implements the main logic (well... at least a placeholder for it) without any knowledge of which particular implementation it is using. Concrete dependencies for each interface are injected into vmmct::app via template parameters.


CMakeLists.txt
```diff
@@ -2,12 +2,12 @@ cmake_minimum_required(VERSION 3.13)
 project(hypervisor CXX ASM-ATT)
 include(${CMAKE_CURRENT_LIST_DIR}/cmake/init_build.cmake)
 
-print_banner()
+hypervisor_print_banner()
```
JaredWright: When you use CMake's FetchContent module and follow the suggested pattern in their documentation, CMake based sub-projects end up behaving as if they are *literally a part* of the project that includes them. This means that CMake functions/macros can trample on each other between projects. For example, PAL includes it's own version of this print_banner CMake function. To allow them all to co-exist nicely, every CMake function needs a globally unique name across hypervisor + all of its dependencies.

cmake/config/default.cmake
```diff
@@ -1,42 +1,42 @@
...
-    CONFIG_TYPE STRING
-    DEFAULT_VAL Release
-    DESCRIPTION "The target build type"
-    OPTIONS Release Debug HelixQAC
-)
-
-add_config(
-    CONFIG_NAME TARGET_ARCH
+hypervisor_add_config(
+    CONFIG_NAME HYPERVISOR_TARGET_ARCH
```
JaredWright: Just like with CMake functions/macros, FetchContent behaves such that subprojects can trample over cache variables and build configurations. For example, PAL has its own TARGET_ARCH configuration that works a little differently than the hypervisor project. Again, the solution is to rename all cache variables and build configs so that they are globally unique.

cmake/depend/bsl.cmake
```diff
@@ -0,0 +1,22 @@
...
+    set(BUILD_EXAMPLES OFF)
+    set(BUILD_TESTS OFF)
+
+    FetchContent_Populate(bsl)
+
+    add_library(bsl INTERFACE)
+    target_include_directories(bsl INTERFACE ${bsl_SOURCE_DIR}/include)
+    # TODO: Use the following instead to add the bsl once it is stable
+    # add_subdirectory(${bsl_SOURCE_DIR} ${bsl_BINARY_DIR})
+endif()
```
JaredWright: Given the limitations with FetchContent described above, there would need to be some CMake updates to the BSL that are very carefully coordinated with the hypervisor project in mind. Specifically, things like target names, cache variables, and functions/macros need to be coordinated.

cmake/depend/pal.cmake
```diff
@@ -0,0 +1,30 @@
...
+        set(PAL_ACCESS_MECHANISM gas_att)
+    elseif(${HYPERVISOR_TARGET_ARCH} STREQUAL armv8a)
+        set(PAL_TARGET_ARCH armv8a)
+        set(PAL_ACCESS_MECHANISM gas_aarch64)
+    endif()
+
+    FetchContent_Populate(pal)
+
+    add_subdirectory(${pal_SOURCE_DIR} ${pal_BINARY_DIR})
+endif()
```
JaredWright: PAL seems to be integrating nicely in to the hypervisor repo using the FetchContent strategy. One thing to keep in mind: adding pal as a dependency to hypervisor comes with a few additional build dependencies

1) You must install Python > 3.6

2) You must ```pip3 install lxml dataclasses colorama pyyaml```

Are we ok with that?

connojd: Glad its working out ok! Personally I'm good with the added dependencies. So going forward any of bareflank's cmake projects will need to consistently use a prefix so they dont clash with others?

JaredWright: Yep, you got the idea! All cache variables and functions could follow a convention like \<PROJECT_NAME\>_\<CONFIG_NAME\>, and that would prevent any collisions from integration via FetchContent.

Even though its a little more verbose, a naming convention like that does make it really easy to glance which configs apply to which projects while viewing in ccmake/cmake-gui. The downside is you end up with quite a few variables that sorta look duplicated (e.g. all the projects have a concept of target_arch, a target_platform, a build_type, etc)

PR #908, opened 2020-05-05
Update build system to include BSL integration

rianquinn:
This patch updates the build system to use the BSL where possible
instead of duplicating things.


cmake/function/hypervisor_add_info.cmake
```diff
@@ -0,0 +1,68 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+include(${bsl_SOURCE_DIR}/cmake/colors.cmake)
```
JaredWright, line 22: There are numerous places like this where the hypervisor's build system is relying on the internal structure of the bsl project, and cmake files that belong to that project. I really think that this is going to end up causing unnecessary complexity, and will do more harm in the long run than the small amount of duplication you are trying to prevent.

example/x64/amd/default/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+    src/main.cpp
+)
+
+target_link_libraries(example_x64_amd_default PRIVATE
+    bsl
+    pal
+    runtime
+)
+
+target_compile_options(example_x64_amd_default PRIVATE
```
JaredWright, line 39: Why did you remove these from the public interface of the runtime CMake target, and instead put them here? Shouldn't the example just inherit these?

example/x64/amd/default/src/start.s
```diff
@@ -0,0 +1,27 @@
...
+/// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+/// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+/// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+/// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+/// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+/// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+/// SOFTWARE.
+
+.global _start
+_start:
```
JaredWright, line 26: Are you intending for every example/extension to define its own _start function? Thats usually something taken care of by a C runtime library, so I would think we would want to do the same pattern here (provide a _start in the runtime project, and let the extension/example define a main function)

cmake/function/hypervisor_add_info.cmake
rianquinn: The hypervisor already is dependent on the internal structure of the BSL. All of the C++ code depends on this. For example: `#include <bsl/array.hypp>`. This is no different. All this means is that the location of CMake APIs is part of the BSL's provided APIs along with the C++ code. If this ever becomes an issue, we can address it then, for now I say we move on.

example/x64/amd/default/CMakeLists.txt
rianquinn: The "-c" and command line warnings are an artifact of the build system. Downstream extensions will almost certainly end up using ExternalProject_Add, in which case these are not needed, and therefore cannot be in the interface. The triple is debatable where it should go. Once we sort out the Clang triple issue, we can figure out where we want these.

example/x64/amd/default/src/start.s
rianquinn: Yes and no. I removed it for now, but it might go back in at some point. Typically, the runtime is actually _start, and the crap around that and nothing more. What's there right now is the syscall logic, which is different. On most platforms, the syscall interface and the CRT are not in the same library. Bareflank currently has a syscall library, a pthread library and a CRT library. pthreads are no longer needed. It might make more sense to rename this to the syscall library, and then create another one specific to the CRT.

example/x64/amd/default/CMakeLists.txt
rianquinn: Ok, I moved the triples to the public interface for now. The "-c" and warnings need to remain private as we don't know how an example will compile itself. This should be fixed

example/x64/amd/default/src/start.s
rianquinn: I renamed it to "syscall" and added a "runtime" that has the start logic in it. I then modified the examples so that they link against both, and the examples no longer have the triple, or their own start.s. If we want to merge the syscall and runtime libraries into a single library in the future we can easily do that, but for now, it makes sense to keep them separate. So... this should be fixed.

rianquinn:
> I can't build this on my mac because the default build is looking for clang-format, which seems to be handled by the bsl. Is there a way for me to configure clang-format off? Seems like that should be optional

Yes, this is optional. You can turn this off using the ENABLE_CLANG_FORMAT option. Once you are using the proper LLVM environment this should not be an issue as Clang Format ships with it. I will also disable this by default for now in the BSL... but be warned that the next PR will enforce Clang Format... so you will need it to make future mods.

rianquinn:
Ok, the BSL has clang format disabled by default.

PR #910, opened 2020-05-07
Adds GitHub Actions and moves to ExternalProject_Add

rianquinn:
This patch does the following:
- This adds in the CI components for automated testing.
  Currently, CodeCov does not work as we do not have tests
  yet, and Perforce has not been tested yet. All other
  build types have been tested on Windows and Linux.
- This moves the build system to ExternalProject_Add.
  Unlike previous attempts at this, our use of ExternalProject_Add
  does not include any magic sauce. We do have a macro, but
  it only forwards specific variables that we want to propagate.
  We also only use ExternalProject_Add once. Meaning, all of
  the cross compiled components are only cross compiled in a
  single step, allowing the PAL and BSL to still use FetchConent,
  and there is no dependency magic, or prefix magic. EPA is
  only used to provide the cross compiled components with
  their own toolchain file which describes how to perform the
  cross compilation.


cmake/toolchain/x86_64.cmake
```diff
@@ -0,0 +1,47 @@
...
+    "-z noexecstack "
+)
+
+set(CMAKE_ASM_COMPILE_OBJECT
+    "<CMAKE_CXX_COMPILER> ${HYPERVISOR_CXX_FLAGS} <DEFINES> <INCLUDES> <FLAGS> -o <OBJECT> -c <SOURCE>")
+
+set(CMAKE_CXX_COMPILE_OBJECT
+    "<CMAKE_CXX_COMPILER> ${HYPERVISOR_CXX_FLAGS} <DEFINES> <INCLUDES> <FLAGS> -o <OBJECT> -c <SOURCE>")
+
+set(CMAKE_CXX_LINK_EXECUTABLE
```
JaredWright, line 45: Why couldn't you have just added this CMAKE_CXX_LINK_EXECUTABLE command to the existing generic toolchain? I feel like you could have left all of these flags at the cmake targets they belong to, updated the --target options where they used to exist, and then not used external project add?

The only thing this toolchain really provides is the ability to link vmm binaries with lld, while not using lld for userspace binaries. Is there some reason that you explicitly don't want to use lld to link the userspace binaries?

rianquinn: If you set the link command it would change linking for everything, including the native compiled parts. The issue with the triple that we need is it will not use the correct linker, and there is no way to change that without either using EPA or making a mod to clang

JaredWright: Right, this sets the link command template for everything in a given cmake project. So what I'm asking is if the following works:
* Remove the HYPERVISOR_CXX_FLAGS and HYPERVISOR_LINK_FLAGS flags from this toolchain file (moving them back to the cmake targets)
* Keep HYPERVISOR_CXX_LINKER in this toolchain file's link command template (which would force cmake to link with lld directly regardless of the clang --target option specified)
* Update the clang --target options to be "more correct" at each individual cmake target (for example, specify --target=x86_64-elf in the kernel and runtime components)

PR #911, opened 2020-05-09
Add VMMCTL and GitIgnore

rianquinn:
This adds the scaffolding for VMMCTL and GitIgnore


loader/linux/src/entry.c
```diff
@@ -0,0 +1,145 @@
...
+
+static int
+resume(void)
+{
+    printk(KERN_INFO "[bareflank_loader]: resume\n");
+    return NOTIFY_DONE;
+}
+
+static int
+suspend(void)
```
connojd, line 84: Can we modify the format rule to use 80 characters until a line break?

loader/linux/src/entry.c
```diff
@@ -0,0 +1,145 @@
...
+}
+
+static struct file_operations fops = {
+    .open = dev_open,
+    .release = dev_release,
+    .unlocked_ioctl = dev_unlocked_ioctl,
+};
+
+static struct miscdevice bareflank_dev = {
+    .minor = MISC_DYNAMIC_MINOR, .name = BAREFLANK_LOADER_NAME, .fops = &fops, .mode = 0666};
```
connojd, line 63: Looks like formatting messed this line up

loader/linux/src/entry.c
```diff
@@ -0,0 +1,145 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * @copyright
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
```
connojd, line 24: The license comments should be changed to GPL since we advertise that with MODULE_LICENSE below

syscall/include/syscall/vcpu.h
```diff
@@ -46,8 +46,7 @@ namespace syscall
     ///   @return Does not return on success. Returns a bsl::errc_type
     ///     on failure.
     ///
-    extern "C" [[nodiscard]] bsl::errc_type
-    run_vcpu(vcpu_context * const vc) noexcept;
+    extern "C" [[nodiscard]] bsl::errc_type run_vcpu(vcpu_context *const vc) noexcept;
```
connojd, line 49: The const is redundant on the right side of the * since vc is on the stack. Did you mean `vcpu_context const *`?

.gitignore
```diff
@@ -56,3 +56,13 @@
 build/
 site/
 doxygen_output/
+
+# Ignore the object files from the Linux loader
+loader/linux/.bareflank_loader.ko.cmd
+loader/linux/.bareflank_loader.mod.o.cmd
+loader/linux/.bareflank_loader.o.cmd
```
connojd, line 63: `loader/linux/.bareflank_loader.mod.cmd` shows up after I build as well

loader/linux/src/entry.c
rianquinn: Not really. We have this configured actually to 100 characters due to the use of namespaces. Clang Format doesn't provide this type of granularity for function prototypes. The way that it is configured is all header prototypes will have the return type on the same line, and all definitions add a newline. There is no way to add any more granularity than that.

loader/linux/src/entry.c
rianquinn: Yeah, I can fix that. It does.

loader/linux/src/entry.c
rianquinn: Nice catch

.gitignore
rianquinn: Yeah, I fixed that as well.

loader/linux/src/entry.c
rianquinn: Ok. I fixed this to use the Dual MIT/GPL license, and then I added the SPDX as required to the linux source. We should be good now to keep the MIT license in there.

PR #922, opened 2020-08-29
Add support for the new MicroV spec

chp-io:
This update provides the needed changes for the new MicroV spec.

### DIFF
--- a/bfvmm/include/hve/arch/intel_x64/ept/mmap.h
+++ b/bfvmm/include/hve/arch/intel_x64/ept/mmap.h
@@ -290,6 +290,186 @@ public:
         return map_4k(reinterpret_cast<void *>(virt_addr), phys_addr, attr, cache);
     }
 
+    /// Share Phys Address Between Two 4k Virt Addresses from a Foreign Mmap
+    ///
+    /// @expects
+    /// @ensures
+    ///
+    /// @return Returns the entry that performs the remap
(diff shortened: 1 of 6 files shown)


bfvmm/include/hve/arch/intel_x64/ept/mmap.h
```diff
@@ -679,6 +825,55 @@ class mmap
 
 private:
 
+    std::pair<std::reference_wrapper<entry_type>, uintptr_t>
+    entry(void *virt_addr, bool should_lock)
+    {
+        if (should_lock) {
+            m_mutex.lock();
+        }
+
```
connojd, line 1056: need to add a gsl::finally call to make sure its unlocked if an exception if thrown before the unlock() call below:
```cpp
if (should_lock) {
    m_mutex.lock();
}

auto unlock = gsl::finally([&]() { if (should_lock) m_mutex.unlock(); });
```

Now when unlock goes out of scope, either through normal return or via exception, the mutex is unlocked. With this you can remove the unlock() call below

chp-io: Thank you for the feedback. I resolved this issue in https://github.com/Bareflank/hypervisor/pull/922/commits/37f0095d263f6cd950223f34f0b46ee55208682d

PR #950, opened 2021-07-14
UEFI: use EFI_LOADED_IMAGE_PROTOCOL to locate boot volume

tkchia:
Some UEFI environments lack `EFI_SHELL_PROTOCOL`, but still implement `EFI_LOADED_IMAGE_PROTOCOL`.

### DIFF
--- a/loader/efi/CMakeLists.txt
+++ b/loader/efi/CMakeLists.txt
@@ -92,6 +92,7 @@ list(APPEND HEADERS
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_file_protocol.h
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_guid.h
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_interface_type.h
+	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_loaded_image_protocol.h
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_locate_search_type.h
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_memory_descriptor.h
 	${CMAKE_CURRENT_LIST_DIR}/include/efi/efi_memory_type.h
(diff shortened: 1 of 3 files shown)


rianquinn:
@tkchia WOW, you did a great job matching the same patterns that we use. Well done. Clang Format is failing. That is really easy to fix. Just ensure that Clang Format is enabled and then run "make format" before submitting and that should fix that specific issue. 

The only actual question I have is, how does this work with respect to which FS it get the files from. Ideally, it would always just load the images sitting next to ./start_bareflank. I have tested on a number of systems, and where it wants to find the kernel and extensions seems to change depending on the system. The code that we put in was a hack to basically just say, always use FS0. Again, would rather that the it just grabbed the images from the same location as to where ever the hypervisor is being started from, but at the very least, if that is not possible, the FS0 hack at least ensures consistency. The mods to this seem to remove the FS0 part of this so my fear is that it will start randomly picking which FS it wants to get the images from like it did before (some would prefer FS0 while others would prefer FS1. 

Or am I just missing something?

tkchia:
Hello @rianquinn,

> Just ensure that Clang Format is enabled and then run "make format" before submitting and that should fix that specific issue.

Thanks!  I have run `ninja format` to reformat the code, and re-pushed the patch.

> The only actual question I have is, how does this work with respect to which FS it get the files from.

My proposed method is to obtain an `EFI_LOADED_IMAGE_PROTOCOL` from the image handle passed to the UEFI entry point, and to use that to get an `EFI_SIMPLE_FILE_SYSTEM_PROTOCOL` &mdash; rather than just to search for any `EFI_SIMPLE_FILE_SYSTEM_PROTOCOL`.

My understanding is that the resulting `EFI_SIMPLE_FILE_SYSTEM_PROTOCOL` should correspond to just the filesystem which the UEFI program was loaded from.  (I guess it might be good to test this, though.)

Thank you!

rianquinn:
@tkchia I will give this some tests and let you know if I have any issues. Thanks a ton for the contributions. Hope to see more of them 👍

tkchia:
Thank you @rianquinn !

